Issue #17, opened 2016-03-18
Take gravity direction as a configurable option


DanielePucci:
The gravity direction used for computing the generalised bias forces is set internally, i.e. 

https://github.com/robotology/WB-Toolbox/blob/master/toolbox/src/InverseDynamics.cpp#L177

It would be desirable to get it as a configurable option.

@traversaro @nunoguedelha @francesco-romano

nunoguedelha:
Maybe that's what you meant, but I think that it would be better if it was retrieved from the IMU measurement, to match Gazebo parameter.

traversaro:
@nunoguedelha The "model" block are not supposed to read data directly , but they provide a flexible input structure such that the user can choose if hardcode the gravity or read it from the IMU.

nunoguedelha:
ok, sure. Thanks

francesco-romano:
Volunteers for testing?

francesco-romano:
Closed by https://github.com/robotology/WB-Toolbox/pull/18

PR #18, opened 2016-03-22
Added explicit gravity option to inverse dynamics

francesco-romano:
As in https://github.com/robotology/WB-Toolbox/issues/17

cc @traversaro @nunoguedelha


francesco-romano:
Just wait before testing.. I discover a new feature (for me)

francesco-romano:
Ready to be tested.
Blocks changed:
- Inverse Dynamics
- Get Bias Forces
- Gravity Compensation

nunoguedelha:
@francesco-romano tested "Get Gravity Forces" and "Inverse Dynamics", comparing both with joint torque measurements:
- for several still positions of the legs
- for two use cases: gravity vector = [0,0,-9.8] or [3,0,-7.8]

... and it's working fine: all three measurements are always equal.

**Note**: for gravity vector = [3,0,-7.8] set in Gazebo, I set [**-3**,0,-7.8] as the simulink blocks input, considering the base frame orientation.

=> didn't test "Get Bias Forces"

francesco-romano:
Great. So I understand I can merge it, right?

nunoguedelha:
Tests on "Get Bias Forces" are ok. Merging.

PR #27, opened 2016-06-23
Added InverseKinematics block

francesco-romano:
Currently this block uses iKin Ipopt-based inverse kinematics


traversaro:
I just realized that we can easily make it more general (arbitrary chains extracted from an URDF) even still using iKin by using the [`iKinLimbFromKDLChain`](http://wiki.icub.org/codyco/dox/html/idyntree/html/group__iDynTreeModelIO.html#gab8ca4dff754c4c5d72e235f5e7396b86). An example of its use (definitely not user-friendly) is given in https://github.com/robotology/idyntree/blob/9e62ff6b886003d697a4b5256fad0e3080f26ce5/src/tools/urdf2dh.cpp#L169 .

francesco-romano:
Because the final aim is to use a new urdf-based inverse kinematics library, I will postpone your suggestions.
Nevertheless, I added the kinematics of r_sole and l_sole, even if a bit hardcoded in the code (the file name is fixed but robot dependent).

PR #28, opened 2016-06-29
Changed inverse kinematics.

francesco-romano:
Now it is possible to specify a starting (base) link and a final (endeffector) link.
cc @traversaro (this should be tested also on linux. I leave this task to @S-Dafarra)


traversaro:
Seems ok. As usual, pay attention to any inconsistency between the limits in the URDF and the one actually returned by the robot in the [`getLimits`](classyarp_1_1dev_1_1IControlLimits) call (the WB-Toolbox of the limits calls the `getLimits` function, while the inverse kinematics block uses the limit from URDF.

francesco-romano:
👍 right.
I'll overwrite the limits passed to the solver by taking the ones from the interface (which can be configured to use URDF or control board limits)

toolbox/CMakeLists.txt
```diff
@@ -244,6 +253,10 @@ else()
     target_link_libraries(WBToolbox ${LINKED_LIBRARIES} ${Matlab_LIBRARIES})
 endif()
 
+# Remote Inverse Kinematics requires C++11
+set(CXX_STANDARD_REQUIRED ON)
+set_property(TARGET WBToolbox PROPERTY CXX_STANDARD 11)
```
traversaro: I suggest to use this:
~~~
if (CMAKE_VERSION VERSION_LESS "3.1")
    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR 
        CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
        target_compile_options(WBToolbox "--std=c++11")
    endif()
else()
    set(CXX_STANDARD_REQUIRED ON)
    set_property(TARGET WBToolbox PROPERTY CXX_STANDARD 11)
endif()
~~~
to preserve compatibility with Ubuntu 14.04 that has CMake 2.8.12 .

traversaro: Eventually (when all the compiler that we support default to use C++11) we will want to drop this to avoid forcing the compiler to use an older language version (for example GCC 6.0 will default to C++14, and I don't know what will happen by forcing it to 11).

francesco-romano: Ok.
Consider that I would have liked to use directly https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html but unfortunately I couldn't find the `mutex` feature

traversaro: Yes that would have been ideal.

traversaro: Apparently is `-std=c++11` with only one dash, sorry for the typo.

francesco-romano: And sorry that I just copied-pasted...

francesco-romano:
@gabrielenava feel free to ask help to @S-Dafarra, so he can also test (again) the inverse kinematics part.

S-Dafarra:
The inverse kinematics part seems ok, same behavior as before.

francesco-romano:
Tested on robot and it seems working

traversaro:
Yeah!

PR #38, opened 2017-01-03
Flexible YARP to DoFs block

francesco-romano:
This PR adds a new block (DoFs converter) which can be used as a mux/demux between YARP serialisation and DoFs serialisation. 
At the current state the block uses the WBI configuration files to detect the WBI serialisation and the YARP one.
It also uses the environment variable `YARP_ROBOT_NAME` to detect the robot.
All the older converters have been removed.

cc @hu-yue my favourite tester 😄  I need feedbacks. Unfortunately I'm not sure this block works perfectly everywhere (because it relies on env variables / matlab variables to be defined), so I need your precious feedback

cc @robotology/codyco-developers if this PR will be merged all the old models won't compile anymore.


francesco-romano:
Tested by @gabrielenava 

Merging

hu-yue:
Nice =D @francesco-romano sorry I saw this now, I was too much into writing.

traversaro:
cc @diegoferigo @S-Dafarra

PR #47, opened 2017-08-10
DiscreteFilter block

diegoferigo:
This block wraps all the filters included in the `iCub::ctrl` namespace ([filters.h](https://github.com/robotology/icub-main/blob/cd2028daac5afcddb220f3b6a7bfe6caa4abdf7b/src/libraries/ctrlLib/include/iCub/ctrl/filters.h)):

* `iCub::ctrl::Filter`
* `iCub::ctrl::FirstOrderLowPassFilter`
* `iCub::ctrl::MedianFilter`

This class already supports multidimensional input signals.
This PR needs the interface that has been discussed in https://github.com/robotology/icub-main/issues/463.

### Future TODOs:

This is what is currently missing and it might be included in future PRs if required:

* Passing filter coefficients during run time (through an additional port)
* Passing initial state from a signal


traversaro:
If we are going to require the `devel` branch of `icub-main` in WB-Toolbox, I think it is convenient to create a `devel` branch also on `icub-main`.

toolbox/src/DiscreteFilter.cpp
```diff
@@ -125,7 +132,10 @@ bool DiscreteFilter::initialize(BlockInformation* blockInfo, wbt::Error* error)
             }
             return 1;
         }
-        filter = new iCub::ctrl::Filter(*num, *den);
+        for (unsigned i = 0; i < inputSignalWidth; ++i) {
```
traversaro: The filter class support vector input/output, why are are you using a different filter for each component of the vector?

toolbox/src/DiscreteFilter.cpp
```diff
@@ -116,6 +118,11 @@ bool DiscreteFilter::initialize(BlockInformation* blockInfo, wbt::Error* error)
     // Create the filter object
     // ========================
 
+    // Get the width of the signals
+    unsigned inputSignalWidth  = blockInfo->getInputPortWidth(INPUT_IDX_SIGNAL);
+    unsigned outputSignalWidth = blockInfo->getOutputPortWidth(OUTPUT_IDX_SIGNAL);
+    assert(inputSignalWidth == outputSignalWidth);
```
traversaro: I am not an expert of Simulink, but what happens with this assert if the user connect two signal of different size to the input and output of the block? Simulink crashes?

toolbox/src/DiscreteFilter.cpp
diegoferigo: With the new commit I forced pushed now the block works with vectors and reduces dynamic allocation

toolbox/src/DiscreteFilter.cpp
diegoferigo: It crashes. In my opinion having asserts is good even in this case. When people work on controllers (downstream) the code should be always set in release. Debug should be enabled only when developing the block themselves. @francesco-romano

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,256 @@
...
+            return true;
+        else
+            return false;
+    };
+
+    // Apply the lambda expression the input parameters
+    s.erase(remove_if(s.begin(), s.end(), lambda_remove_chars), s.end());
+
+    // Convert the cleaned string to a yarp vector of floats
+    std::istringstream sstrm(s);
```
traversaro: Pay attention, as this conversion part could be affected by locale-bugs such as https://github.com/robotology/idyntree/issues/288 .

diegoferigo: Fixed in https://github.com/robotology/WB-Toolbox/pull/47/commits/aafaf3873df806f8bf454b720c815779142e70e4

toolbox/src/base/factory.cpp
```diff
@@ -36,7 +36,10 @@ wbt::Block* wbt::Block::instantiateBlockWithClassName(std::string blockClassName
 #ifdef WBT_USES_ICUB
     else if (blockClassName == wbt::MinimumJerkTrajectoryGenerator::ClassName) {
         block = new wbt::MinimumJerkTrajectoryGenerator();
-    } 
+    }
+    else if (blockClassName == wbt::DiscreteFilter::ClassName) {
```
francesco-romano: Please keep the indentation of the else coherent with the rest of the file

diegoferigo: What do you mean? It looks already coherent but is morning for me as well :)

toolbox/include/DiscreteFilter.h
```diff
@@ -0,0 +1,47 @@
...
+    namespace sig {
+        class Vector;
+    }
+} // namespace yarp
+
+class wbt::DiscreteFilter : public wbt::Block {
+private:
+    bool firstRun;
+    iCub::ctrl::IFilter* filter;
+    yarp::sig::Vector* num;
```
francesco-romano: Why putting `num` and `den` in the `.h` file? They are only needed in configuration and not in real time.

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+using namespace wbt;
+using namespace iCub::ctrl;
+using namespace yarp::sig;
+
+std::string DiscreteFilter::ClassName = "DiscreteFilter";
+
+DiscreteFilter::DiscreteFilter() : firstRun(true), filter(nullptr), inputSignalVector(nullptr)
+{
+    num = new Vector(0);
```
francesco-romano: see comment in the `.h`

toolbox/src/base/factory.cpp
diegoferigo: Oh sorry, now I see. Inside `WBT_USES_ICUB` is coherent, outside no. I fix both.

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,261 @@
...
+    if (filter == nullptr) return false;
+
+    // Get the input and output signals
+    Signal inputSignal  = blockInfo->getInputPortSignal(INPUT_IDX_SIGNAL);
+    Signal outputSignal = blockInfo->getOutputPortSignal(OUTPUT_IDX_SIGNAL);
+
+    unsigned inputSignalWidth = blockInfo->getInputPortWidth(INPUT_IDX_SIGNAL);
+
+    // Allocate the memory for the input data
+    if (firstRun) {
```
francesco-romano: There is also another callback (`initializeInitialConditions`) which is called before the first output.

Furthermore, have you tried to ask for the port width in the `initialize` method?

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+    // Forward the filtered signals to the output port
+    outputSignal.setBuffer(outputVector.data(), outputVector.length());
+
+    return true;
+}
+
+void DiscreteFilter::stringToYarpVector(const std::string str, Vector* v)
+{
+    assert(v != nullptr);
```
francesco-romano: As the pointer should always be `!=` null, I would use a reference, not a pointer

toolbox/src/base/factory.cpp
diegoferigo: Mmh consideing that the last part of the file is all with defines, and the indentation has to be changes, imo the current status looks better. Compare:
```cpp
    ...
    } else if (blockClassName == wbt::GetLimits::ClassName) {
        block = new wbt::GetLimits();
    } else if (blockClassName == wbt::CentroidalMomentum::ClassName) {
        block = new wbt::CentroidalMomentum();
    }
#ifdef WBT_USES_ICUB
    else if (blockClassName == wbt::MinimumJerkTrajectoryGenerator::ClassName) {
        block = new wbt::MinimumJerkTrajectoryGenerator();
    }
    else if (blockClassName == wbt::DiscreteFilter::ClassName) {
        block = new wbt::DiscreteFilter();
    }
#endif
#ifdef WBT_USES_IPOPT
    else if (blockClassName == wbt::InverseKinematics::ClassName) {
        block = new wbt::InverseKinematics();
    }
#endif
```
with
```cpp
    ...
    } else if (blockClassName == wbt::GetLimits::ClassName) {
        block = new wbt::GetLimits();
    } else if (blockClassName == wbt::CentroidalMomentum::ClassName) {
        block = new wbt::CentroidalMomentum();
    }
#ifdef WBT_USES_ICUB
    else if (blockClassName == wbt::MinimumJerkTrajectoryGenerator::ClassName) {
        block = new wbt::MinimumJerkTrajectoryGenerator();
    } else if (blockClassName == wbt::DiscreteFilter::ClassName) {
        block = new wbt::DiscreteFilter();
    }
#endif
#ifdef WBT_USES_IPOPT
    else if (blockClassName == wbt::InverseKinematics::ClassName) {
        block = new wbt::InverseKinematics();
    }
#endif
```

francesco-romano: I prefer the second one.

toolbox/src/DiscreteFilter.cpp
diegoferigo: In `initialize()` it returns the correct width. I'll add a private member for that and move the initialization into the `initializeInitialConditions` callback that I didn't know existed.

francesco-romano: if `initialize` returns the correct width, I advise to use it instead of `initializeInitialConditions`

toolbox/include/DiscreteFilter.h
diegoferigo: Done

toolbox/src/DiscreteFilter.cpp
diegoferigo: Done

toolbox/src/DiscreteFilter.cpp
diegoferigo: Ok thanks, should be ok now.

toolbox/src/DiscreteFilter.cpp
diegoferigo: Changed the argument type to `Vector&`

diegoferigo:
@francesco-romano 
Ready for the final review

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,317 @@
...
+}
+
+bool DiscreteFilter::initializeInitialConditions(BlockInformation* blockInfo, wbt::Error* error)
+{
+    // Reminder: this function is called when, during runtime, a block is disabled
+    // and enabled again. The method ::initialize instead is called just one time.
+
+    // If the initial conditions for the output are not set, allocate a properly
+    // sized vector
+    if (*y0 == Vector(0)) {
```
francesco-romano: Why not asking for the size of the vector and calling a resize on it, instead of allocated twice?

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+using namespace wbt;
+using namespace iCub::ctrl;
+using namespace yarp::sig;
+
+std::string DiscreteFilter::ClassName = "DiscreteFilter";
+
+DiscreteFilter::DiscreteFilter() : filter(nullptr), inputSignalVector(nullptr)
+{
+    y0 = new Vector(0);
```
francesco-romano: As the delete are done in the terminate, move the allocation in the corresponding method, and not in the constructor.
Here simply initialise to `nullptr` the pointers

toolbox/src/DiscreteFilter.cpp
```diff
@@ -0,0 +1,317 @@
...
+        y0 = new Vector(inputSignalWidth, 0.0);
+    }
+
+    // Initialize the filter. This is required because if the signal is not 1D,
+    // the default filter constructor initialize a wrongly sized y0
+    // Moreover, the Filter class has a different constructor that handles the
+    // zero-gain case
+    Filter* filter_c = dynamic_cast<Filter*>(filter);
+    if (filter_c != nullptr) {
+        filter_c->init(*y0, *u0);
```
francesco-romano: I don't remember what `init` does, but is it ok to call with an empty vector? (i.e. `u0.size() == 0`)

diegoferigo: With https://github.com/robotology/WB-Toolbox/pull/47/commits/4a292316a247fb014992796aa8578580604dff12 the vectors that hold the optional initial values are null if the parameters are not specified, and allocated to the correct size only when needed. If instead they are specified, their size is checked in the `initialize` method.

toolbox/src/DiscreteFilter.cpp
```diff
@@ -246,8 +249,12 @@ bool DiscreteFilter::initializeInitialConditions(BlockInformation* blockInfo, wb
 
     // If the initial conditions for the output are not set, allocate a properly
     // sized vector
-    if (*y0 == Vector(0)) {
-        y0 = new Vector(inputSignalWidth, 0.0);
+    if (y0 == nullptr) {
+        unsigned outputSignalWidth = blockInfo->getInputPortWidth(OUTPUT_IDX_SIGNAL);
+        y0                         = new Vector(outputSignalWidth, 0.0);
```
francesco-romano: can you avoid to align the `=`? I don't like this kind of indentation

francesco-romano:
A part from the inline comment the `.cpp` is ok.
As soon as I have time I will test the `.mdl`

francesco-romano:
@diegoferigo 
To "rebase" on top of devel, rebase the C++ files. 
Regarding the mdl, save the new block in a temporary model, get the devel version of the library and then re-add the block

diegoferigo:
For simplifying the test of the WB-Toolbox 3.0, I cherry picked these commits to https://github.com/robotology/WB-Toolbox/pull/56.

Closing this PR.

PR #54, opened 2017-09-05
Get Control references block

francesco-romano:


francesco-romano:
@S-Dafarra if this is ok, we can merge.

S-Dafarra:
It is ok for me!

S-Dafarra:
It seems to be very slow when used on iCub.

diegoferigo:
What is the current status of this PR? Does it have to rebased and merged, or closed?

cc @S-Dafarra @francesco-romano

francesco-romano:
I don't know. I will let @S-Dafarra comment on this.

S-Dafarra:
As far as I remember the idea was to not merge it on ``WBToolbox 2.0`` given that the new version was about to be released. This block was used to get the references given to the low level controllers (position references, velocities, torques,...). Despite the urgency behind this block has consistently reduced, I think it can be an useful debugging tool also for controllers which are not written in Simulink. 

Just as a side note. If we can write this block in the 3.0 version, it would be better to use the ``getReferences`` from the ``iPidControl`` interface.

diegoferigo:
@S-Dafarra Thanks for the explanation. I'm going to close this PR. I just opened #54 as a reminder of this feature.

Issue #53, opened 2017-08-23
Configure library without yarpWholeBodyInterface.ini file


francesco-romano:

diegoferigo:
Implemented in #56

Issue #52, opened 2017-08-23
Substitute yarpWholeBodyActuators by interfacing to Yarp controlboards


francesco-romano:

diegoferigo:
Implemented in #56

Issue #51, opened 2017-08-23
Substitute yarpWholeBodyModel with iDynTree


francesco-romano:

diegoferigo:
Implemented in #56

Issue #50, opened 2017-08-23
Directly obtain joint velocities from control board


francesco-romano:
Get the velocities from the ControlBoardRemapper. The new robots have now encoders and it is not needed estimating them

diegoferigo:
Implemented in #56

PR #56, opened 2017-09-11
WB-Toolbox 3.0

diegoferigo:
Refer to https://github.com/robotology/WB-Toolbox/issues/46 and [this milestone](https://github.com/robotology/WB-Toolbox/milestone/2) for the task list. 

This PR:
* Fixes https://github.com/robotology/WB-Toolbox/issues/53 Configure library without yarpWholeBodyInterface.ini file
* Fixes https://github.com/robotology/WB-Toolbox/issues/52 Substitute yarpWholeBodyActuators by interfacing to Yarp controlboards
* Fixes https://github.com/robotology/WB-Toolbox/issues/51 Substitute yarpWholeBodyModel with iDynTree
* Fixes https://github.com/robotology/WB-Toolbox/issues/50 Directly obtain joint velocities from control board
* Fixed https://github.com/robotology/WB-Toolbox/issues/37 Multiple robots in the same simulink model

**`[WIP]`**

- [x] Update all the Simulink Masks
- [x] `WBTPIDConfig` still misses a `get` method returning a struct
- [x] Move other matlab scripts into the `+WBToolbox` package
- [x] Somewhere `std::vector::assign()` is used on non non-contiguous arrays, double check all blocks
- [x] Convert all remaining raw pointers to smart pointers
- [ ] Improve doxygen documentation
- [x] Check if `rad2deg` is required in `SetReferences.cpp`
- [x] Double check buffered vector in `YarpRead.cpp`
- [x] Remove the `wbt::` namespace nesting in the `cpp` files
- [x] Add warnings messages. Right now this new feature is still rarely used
- [x] Evaluate if removing Eigen from headers
- [x] Write a Changelog
- [x] Update `README.md`
- [ ] Write documentation


toolbox/include/CentroidalMomentum.h
```diff
@@ -2,21 +2,30 @@
...
-    double *m_basePose;
-    double *m_centroidalMomentum;
+class wbt::CentroidalMomentum : public wbt::WBIModelBlock
+{
+    // Robot state
+    iDynTree::Twist*         m_baseVelocity;
+    iDynTree::Vector3*       m_gravity;
+    iDynTree::Transform*     m_world_T_base;
+    iDynTree::VectorDynSize* m_jointsVelocity;
+    iDynTree::VectorDynSize* m_jointsState;
```
traversaro: In mechanical systems, "state" is tipically used to indicate position ***and*** velocity, so I would call this member `m_jointsPosition` or something similar.

diegoferigo: I didn't like the previous word *configuration* and I switched to *State*. *Position* is even more clear :+1:

toolbox/include/DotJNu.h
```diff
@@ -2,23 +2,33 @@
...
 namespace wbt {
     class DotJNu;
 }
 
-class wbt::DotJNu : public wbt::WBIModelBlock {
+namespace iDynTree {
+    class Twist;
+    class Transform;
+    class VectorDynSize;
+    typedef int FrameIndex;
```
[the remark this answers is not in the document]
diegoferigo: I tried to keep as much as possible the same behavior as the current state, and I agree with you on these forward declaration. For `VectorFixSize` I had to add the header since redefining templates was a bit too much imo, but applying this line to all the typedefs would make sense too

toolbox/src/CentroidalMomentum.cpp
```diff
@@ -4,154 +4,203 @@
...
         unsigned dofs = WBInterface::sharedInstance().numberOfDoFs();
-        m_basePose = new double[16];
-        m_centroidalMomentum = new double[6];
-        m_basePoseRaw = new double[16];
-        m_configuration = new double[dofs];
-        m_baseVelocity = new double[6];
-        m_jointsVelocity = new double[dofs];
-
-        return m_basePose && m_centroidalMomentum && m_basePoseRaw && m_configuration && m_baseVelocity && m_jointsVelocity;
+        double m_gravityRaw[3] {0, 0, -9.81};
```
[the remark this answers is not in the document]
diegoferigo: So far the only block with an optional input for gravity is [`InverseDynamics`](https://github.com/diegoferigo/WB-Toolbox/blob/82b48feb52ef097a71c53029e8765ba227b61795/toolbox/src/InverseDynamics.cpp#L247). So far I tried to maintain the current functionalities, without the need of editing the Simulink's masks. I'll add this in a future TODO list

toolbox/src/ForwardKinematics.cpp
```diff
@@ -4,156 +4,230 @@
...
             if (error) error->message = "Failed to configure the number of input ports";
             return false;
         }
+
+        // Size and type
         bool success = true;
-        success = success && blockInfo->setInputPortMatrixSize(0, 4, 4);
-        success = success && blockInfo->setInputPortVectorSize(1, dofs);
+        success &= blockInfo->setInputPortMatrixSize(INPUT_IDX_BASE_POSE, 4, 4);
+        success &= blockInfo->setInputPortVectorSize(INPUT_IDX_JOINTCONF, dofs);
```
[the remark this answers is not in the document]
diegoferigo: I found the previous version redundant, if @francesco-romano has no comments on this I'd like to use the new one.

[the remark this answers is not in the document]
francesco-romano: honestly I am against the use of `&=`

toolbox/src/DotJNu.cpp
```diff
@@ -4,181 +4,253 @@
...
+            if (m_frameIndex == FRAME_INVALID_INDEX) {
                 if (error) error->message = "Cannot find " + frame + " in the frame list";
                 return false;
             }
-        } else {
-            m_frameIndex = wbi::wholeBodyInterface::COM_LINK_ID;
         }
-        
+        else {
+            m_frameIndex = COM_LINK_ID;
```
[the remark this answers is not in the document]
diegoferigo: @traversaro In the `output()` method of all the blocks that use `COM_LINK_ID` the correct iDynTree's method for CoM is called. Since all integers `<0` are treated as FRAME_INVALID_INDEX by iDynTree, do you have any suggestion on how to handle this case with a viarable?

I mean, when `m_frameIndex = -1` and the frame is `com`, an error is thrown. If `m_frameIndex = -1` and the frame is not `com`, the block goes further to the `output()` and the `COM_LINK_ID` is selected [to chose the alternative method](https://github.com/robotology/WB-Toolbox/pull/56/files/82b48feb52ef097a71c53029e8765ba227b61795#diff-ff0898c2cbf4dce4de854b3e206599bdR241) for computations on CoM.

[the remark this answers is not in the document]
traversaro: I see, I missed the check on `COM_LINK_ID` . Everything should work fine indeed, but personally I would prefer to have a separated boolean variable to save the information "is this block computing COM-related stuff", rather than overloading the meaning of `m_frameIndex` . However, this is mostly a personal taste, so feel free to use the pattern that you prefer. The only advice that I would recommend is to define  `COM_LINK_ID` to any negative number different from `-1` , as `-1` is exactly the value of `iDynTree::FRAME_INVALID_INDEX`.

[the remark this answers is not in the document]
francesco-romano: I like the boolean variable idea. It is the most safe and clean in the long run

diegoferigo:
> Do you think it would make sense to add an explicitly call to setFrameVelocityRepresentation to make sure that the representation is the intended one?

@traversaro Can you please give me some further detail on this? What representation are we using inside the Toolbox? I'm a bit confused on this.

diegoferigo:
@francesco-romano @traversaro 
First draft of the `WB-Toolbox 3.0` pushed. For the time being only C++.

toolbox/include/base/BlockInformation.h
```diff
@@ -59,19 +60,22 @@ class wbt::BlockInformation {
...
+    // =====================
 
     /**
      * Convert a block option from its Toolbox identifier to a specific implementation
      *
-     * @param [in] key identifier of the block option
-     * @param [out] option implementation-specific block option
-     * @return true if the option has been converted. False otherwise
+     * @param [in] key    identifier of the block option
+     * @param [in] option implementation-specific block option
```
traversaro: Is this an in or out parameter?

francesco-romano:
I think we use Mixed representation, am I right @traversaro ?

toolbox/CMakeLists.txt
```diff
@@ -275,6 +273,10 @@ else()
     target_link_libraries(WBToolbox ${LINKED_LIBRARIES} ${Matlab_LIBRARIES})
 endif()
 
+# Link with ClockServer library
+add_subdirectory(autogenerated/)
+target_link_libraries(WBToolbox PUBLIC ClockServer)
```
francesco-romano: Move these two lines next to the Synchronizer block, so in the future you can `if` the whole block

toolbox/autogenerated/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+cmake_policy(SET CMP0048 NEW)
+project(ClockServer LANGUAGES CXX VERSION 0.1)
```
francesco-romano: I don't think it is needed to create different projects. A target should be sufficient (? @traversaro ?)

toolbox/src/base/Signal.cpp
```diff
@@ -223,19 +277,19 @@ namespace wbt {
...
                 dataSize = sizeof(uint32_t);
-                address = ((uint32_t*)data) + startIndex;
+                address = data + startIndex;
                 break;
             }
             case PortDataTypeUInt16:
             {
                 dataSize = sizeof(uint16_t);
-                address = ((uint16_t*)data) + startIndex;
+                address = data + startIndex;
```
francesco-romano: Why did you drop the cast?

toolbox/src/base/Signal.cpp
```diff
@@ -223,19 +277,19 @@ namespace wbt {
...
                 break;
             }
             case PortDataTypeUInt8:
             {
                 dataSize = sizeof(uint8_t);
-                address = ((uint8_t*)data) + startIndex;
+                address = data + startIndex;
```
francesco-romano: Why did you drop the cast?

toolbox/include/base/BlockInformation.h
```diff
@@ -68,10 +70,12 @@ class wbt::BlockInformation {
      * @param [out] option implementation-specific block option
      * @return true if the option has been converted. False otherwise
      */
-    virtual bool optionFromKey(const std::string& key, Data &option) const;
+    virtual bool optionFromKey(const std::string& key, double& option) const;
```
francesco-romano: Why assuming that options can be only double?

toolbox/include/base/WBBlock.h
```diff
@@ -0,0 +1,79 @@
...
+ * you probabily want to derive from Block instead)
+ */
+class wbt::WBBlock : public wbt::Block
+{
+private:
+    static const unsigned PARAM_IDX_CONFIGURATION;
+    static const unsigned PARAM_IDX_CONFBLOCKNAME;
+
+protected:
+    std::string confKey;
```
francesco-romano: `configurationID` ?

toolbox/include/base/WBBlock.h
```diff
@@ -0,0 +1,79 @@
...
+ * - During terminate it closes and release the interface object and terminate the yarp network
+ *
+ * @note Usually you want to call this class implementations at some point in your
+ * method overridings, unless you want to completely change the code (but at that point
+ * you probabily want to derive from Block instead)
+ */
+class wbt::WBBlock : public wbt::Block
+{
+private:
+    static const unsigned PARAM_IDX_CONFIGURATION;
```
francesco-romano: At this point, completely change the name:
`ConfigurationParameterIndex`

toolbox/include/base/WBBlock.h
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * \struct iDynTreeRobotState WBBlock.h
+ *
+ * This struct contains the pointer to iDynTree objects used to configure the
+ * state of iDynTree::KinDynComputations objects.
+ */
+struct iDynTreeRobotState {
+    iDynTree::Twist*         m_baseVelocity   = nullptr;
```
francesco-romano: Do we really need these to be pointers?

I would also add a constructor (with dofs as parameter) and destructor to simplify memory management

toolbox/src/base/WBBlock.cpp
```diff
@@ -0,0 +1,275 @@
...
+    robotState.m_baseVelocity   = new Twist();
+    robotState.m_jointsVelocity = new VectorDynSize(dofs);
+    robotState.m_jointsVelocity->zero();
+    return robotState.m_gravity && robotState.m_jointsPosition && robotState.m_world_T_base &&
+           robotState.m_baseVelocity && robotState.m_jointsVelocity;
+}
+
+unsigned WBBlock::numberOfParameters() { return 2; }
+
+bool WBBlock::getWBToolboxParameters(Configuration& config, BlockInformation* blockInfo)
```
francesco-romano: I probably forgot at the beginning to add `const` specifier to the blockInfo parameter (in the whole toolbox. It does not make sense to modify it once created by Simulink)

toolbox/src/base/WBBlock.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    // Check if the configuration is valid
+    if (!config.isValid()) {
+        Log::getSingleton().Error("The provided configuration is not valid.");
+        return false;
+    }
+
+    // Store the configuration inside the ToolboxSingleton
+    if (!interface.configure(confKey, config)) {
```
francesco-romano: I don't like `configure` as it seems it configure something, while actually it is just saving the configuration, am I right?

toolbox/src/base/WBBlock.cpp
```diff
@@ -0,0 +1,275 @@
...
+        delete robotState.m_baseVelocity;
+        robotState.m_baseVelocity = nullptr;
+    }
+    if (robotState.m_jointsVelocity) {
+        delete robotState.m_jointsVelocity;
+        robotState.m_jointsVelocity = nullptr;
+    }
+
+    // If the RemoteControlBoardRemapper has been allocated, free its memory
+    ToolboxSingleton& interface = ToolboxSingleton::sharedInstance();
```
francesco-romano: I think that this part should be managed by the specific subclasses.
If we do not explicitly `retain` in the initialise we must not call `release` in the terminate

toolbox/include/base/Block.h
```diff
@@ -15,14 +14,15 @@ namespace wbt {
  * Basic abstract class for all the blocks.
  * This class (the whole toolbox in reality) assumes the block represent
  * an instantaneous system (i.e. not a dynamic system).
- * 
+ *
  * You can create a new block by deriving this class and implementing at least
  * all the pure virtual methods.
  *
  * @Note: if you need to implement a block which uses the WBI, you should derive
```
francesco-romano: Document that if a method returns false, one has to retrieve the singleton Log object

toolbox/src/base/Signal.cpp
```diff
@@ -237,7 +234,7 @@ namespace wbt {
                 break;
         }
 
-        memcpy(contiguousData, address, dataSize * length);
+        memcpy(contiguousData, address, dataSize* length);
```
francesco-romano: revert this change. 
binary operators should have a space both before and after

toolbox/src/base/Signal.cpp
```diff
@@ -296,22 +293,22 @@ namespace wbt {
...
```
francesco-romano: same as before

toolbox/src/base/Signal.cpp
```diff
@@ -296,22 +293,22 @@ namespace wbt {
...
 
-    void Signal::setBuffer(const bool *data, const unsigned length, unsigned startIndex)
+    void Signal::setBuffer(const bool* data, const unsigned length, unsigned startIndex)
     {
         if (isConstPort) return;
         unsigned dataSize = sizeof(bool);
         const void* address = static_cast<const bool*>(data) + startIndex;
 
-        memcpy(contiguousData, address, dataSize * length);
+        memcpy(contiguousData, address, dataSize* length);
```
francesco-romano: same as before

toolbox/include/base/ToolboxSingleton.h
```diff
@@ -0,0 +1,142 @@
...
+    /// @see wbt::RobotInterface
+    std::unordered_map<std::string, std::shared_ptr<wbt::RobotInterface>> m_interfaces;
+
+public:
+
+    // CONSTRUCTOR / DESTRUCTOR
+    // ========================
+
+    ToolboxSingleton();
+    ToolboxSingleton(const ToolboxSingleton&);
```
francesco-romano: If it is a singleton both copy and assignment should be prohibited

toolbox/include/base/ToolboxSingleton.h
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Returns a \c shared_ptr to the KinDynComputations object used to perform calculation
+     * on the provided model
+     *
+     * @param confKey The key describing the configuration (name of the Simulink block)
+     * @return        A \c shared_ptr to the iDynTree::KinDynComputations of the requested
+     *                configuration
+     */
+    const std::shared_ptr<iDynTree::KinDynComputations> model(const std::string& confKey);
```
francesco-romano: To be consistent with the previous two methods I would rename this into `getModel`

toolbox/include/base/ToolboxSingleton.h
```diff
@@ -0,0 +1,142 @@
...
+     * @note Since confKey is the name of the block, the overriding can happen only after
+     *       subsequent compilation in Simulink, when the block's parameters have been
+     *       changed.
+     *
+     * @param  confKey The key describing the configuration (name of the Simulink block)
+     * @param  config  The wbt::Configuration object parsed from Simulink's parameters
+     * @return         Returns \c true if configure is successful, \c false otherwise
+     * @see            ToolboxSingleton::isKeyValid
+     */
+    bool configure(const std::string& confKey, const Configuration& config);
```
francesco-romano: as before, I do not like the name `configure`

toolbox/src/base/ToolboxSingleton.cpp
```diff
@@ -0,0 +1,124 @@
...
+        if (!config.isValid()) {
+            return false;
+        }
+
+        // Add the new Configuration object and override an existing key if it already exist.
+        // Note: Simulink doesn't flush memory unless Matlab is closed, and static objects stay in memory.
+        //       This may cause problems if the config block's mask is changed after the first compilation.
+        if (m_interfaces.find(confKey) == m_interfaces.end()) {
+            m_interfaces[confKey] = std::make_shared<RobotInterface>(config);
+            return m_interfaces[confKey].get();
```
francesco-romano: This comment applies also to other uses of `.get()`.
`shared_ptr` have an `operator bool`. This means that you can directly return `return m_interfaces[confKey]` to assess validity of the pointer.

Change this line and the other times you use `get()` as a boolean check

toolbox/include/base/BlockInformation.h
diegoferigo: An out parameter, I don't even remember why I changed this

+WBToolbox/WBToolboxConfig.m
```diff
@@ -0,0 +1,156 @@
+classdef WBToolboxConfig < matlab.mixin.Copyable
+    %WBTOOLBOXCONFIG Summary of this class goes here
+    %   Detailed explanation goes here
+
+    % PROPERTIES
+    % ==========
+
+    properties
+        % Robot state
+        RobotName          (1,:) char
```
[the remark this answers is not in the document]
diegoferigo: [Yes](https://it.mathworks.com/help/stateflow/api/accessing-the-properties-and-methods-of-objects.html#f13-5782)

+WBToolbox/WBToolboxConfig.m
```diff
@@ -0,0 +1,156 @@
...
+
+    methods
+
+        % SET METHODS
+        % -----------
+
+        % Dependent properties
+        % --------------------
+
+        function set.ValidConfiguration(~,~)
```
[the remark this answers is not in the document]
diegoferigo: `ValidConfiguration` is a `Dependent` property, so by default even without this set method an error is generated. Specifying this method gives the possibility to provide a custom error message, and in this case it is used exactly for this reason.

+WBToolbox/WBToolboxConfig.m
```diff
@@ -0,0 +1,156 @@
...
+        function value = get.ValidConfiguration(obj)
+            % Check if all the properties have been set
+            value = ...
+                ~isempty(obj.RobotName) && ...
+                ~isempty(obj.UrdfFile)  && ...
+                ~isempty(obj.ControlledJoints)   && ...
+                ~isempty(obj.ControlBoardsNames) && ...
+                ~isempty(obj.LocalName) && ...
+                ~isequal(obj.GravityVector,[0, 0, 0]);
+            % Check if the urdf file has the right extension
```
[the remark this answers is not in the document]
diegoferigo: Ok, let's be generic then

MxAnyType/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+cmake_policy(SET CMP0048 NEW)
+project(MxAnyType LANGUAGES CXX VERSION 0.1)
+cmake_minimum_required(VERSION 3.0.2)
+
+# Configure the project
+# =====================
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include(GNUInstallDirs)
+option(BUILD_SHARED_LIBS OFF)
```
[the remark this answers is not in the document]
diegoferigo: Removed

MxAnyType/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+
+
+# TODO: temporary, waiting the library becomes a shared
+if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+    set_target_properties(${VARS_PREFIX} PROPERTIES COMPILE_FLAGS "-fPIC")
+endif()
+
+# Find Eigen
+find_package(Eigen3 REQUIRED)
+target_include_directories(${VARS_PREFIX} PUBLIC "${EIGEN3_INCLUDE_DIR}")
```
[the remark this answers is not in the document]
diegoferigo: Ping @traversaro. Has Eigen been fixed on Windows when used in headers? Any comment / alternatives on this?

MxAnyType/include/MxAnyType.h
```diff
@@ -0,0 +1,99 @@
...
+class MxAnyType;
+
+// If needed in the future
+// class MxAnyCell : public AnyCell {};
+// class MxAnyStruct : public AnyStruct {};
+
+struct MxArrayMetadata {
+    mxClassID id;
+    bool isScalar;
+    unsigned rows;
```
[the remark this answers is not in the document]
diegoferigo: Changed to `size_t`

MxAnyType/src/MxAnyType.cpp
```diff
@@ -0,0 +1,219 @@
...
+// STRING / CHARS
+// ==============
+
+bool MxAnyType::asString(std::string& s)
+{
+    if (!mx) return false;
+    if (md.id != mxCHAR_CLASS) return false;
+    char* buffer = mxArrayToString(mx);
+    s = std::string(buffer);
+    return true;
```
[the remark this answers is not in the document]
diegoferigo: Yes, this is one of the few cases the user has to handle memory

MxAnyType/src/MxAnyType.cpp
```diff
@@ -0,0 +1,219 @@
...
+bool MxAnyType::asAnyStruct(AnyStruct& s)
+{
+    if (!mx) return false;
+    if (md.id != mxSTRUCT_CLASS) return false;
+
+    // MxAnyStruct& mxStruct = dynamic_cast<MxAnyStruct&> s;
+    // if (!s) return false;
+
+    for (unsigned i = 0; i < mxGetNumberOfFields(mx); ++i) {
+        const char* fieldName = mxGetFieldNameByNumber(mx,i);
```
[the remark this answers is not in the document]
diegoferigo: These small details will be handled with the very last commit which will apply a style with `clang-format`

MxAnyType/src/MxAnyType.cpp
```diff
@@ -0,0 +1,219 @@
...
+        // TODO multidimensional struct
+        mxArray* fieldContent = mxGetFieldByNumber(mx,0,i);
+        if (fieldName == nullptr) return false;
+        if (fieldContent == nullptr) return false;
+        // TODO who should delete this memory? AnyStruct?
```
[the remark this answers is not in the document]
diegoferigo: As per bullet list in the first post, I'd like to avoid as much as possible raw pointers. Switching to smart pointers is on the TODO list.

MxAnyType/src/MxAnyType.cpp
```diff
@@ -0,0 +1,219 @@
...
+        MxAnyType* fieldContentMxAnyType = new MxAnyType(fieldContent);
+        AnyType* fieldContentAnyType = static_cast<AnyType*>(fieldContentMxAnyType);
```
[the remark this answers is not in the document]
diegoferigo: Substituted these two portion of code with a single line:
```
s[std::string(fieldName)] = std::make_shared<MxAnyType>(fieldContent);
```

MxAnyType/src/MxAnyType.cpp
```diff
@@ -0,0 +1,219 @@
...
+{
+    if (!mx) return false;
+    if (md.id != mxCELL_CLASS) return false;
+
+    // TODO: AnyCell then will have a operator()(3,4) method;
+    for (unsigned i=0; i < mxGetNumberOfElements(mx); ++i) {
+        mxArray* cellContent = mxGetCell(mx, i);
+        if (!cellContent) return false;
+        MxAnyType* cellContentMxAnyType = new MxAnyType(cellContent);
+        AnyType* cellContentAnyType = static_cast<AnyType*>(cellContentMxAnyType);
```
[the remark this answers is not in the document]
diegoferigo: As above

toolbox/CMakeLists.txt
diegoferigo: It is not possible moving this portion of code before the target `WBToolbox` is created

toolbox/autogenerated/CMakeLists.txt
diegoferigo: Here I preferred starting from scratch with a new project in order to have them as much independent as possible. Suggestions are welcome.

toolbox/include/base/BlockInformation.h
diegoferigo: This was the first step of deprecating the `Data` class. "Was" because I forgot it is used also for `Signal`s. Internally in any case the value is saved in a double format, even if it is a bool.

After this new WB-Toolbox version be properly tested, I'd like to improve `Data` as described [here](https://github.com/robotology/WB-Toolbox/issues/60). Now there are too many changes to test, and I'm postponing activities.

toolbox/include/base/Configuration.h
```diff
@@ -0,0 +1,162 @@
...
+class wbt::Configuration
+{
+// TODO: check how localName is used
+private:
+    std::string m_robotName; ///< Name of the robot
+    std::string m_urdfFile;  ///< Name of the file containing the urdf model
+    std::string m_localName; ///< Prefix appended to the opened ports
+    std::vector<std::string> m_controlledJoints;   ///< Subset of controlled joints
+    std::vector<std::string> m_controlBoardsNames; ///< Names of the used ControlBoard names
+    std::vector<double> m_gravityVector; ///< The gravity vector
```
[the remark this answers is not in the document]
diegoferigo: I don't like so much C arrays in C++ if not strictly necessary. My first idea was to use a `std::array`, but it is not possible to extract the raw array from it. Unfortunately somewhere we need this feature.

toolbox/include/base/Configuration.h
```diff
@@ -0,0 +1,162 @@
...
+    unsigned m_dofs; //< DoFs extracted my m_controlBoardsNames vector
```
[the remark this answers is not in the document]
diegoferigo: :+1:

toolbox/include/base/Configuration.h
```diff
@@ -0,0 +1,162 @@
...
+                       std::vector<std::string> controlBoardsNames,
+                       std::string localName,
+                       std::vector<double> gravityVector);
+
+    /**
+     * Set the name of the robot
+     *
+     * @param robotName Name of the robot
+     */
+    void setRobotName(std::string robotName);
```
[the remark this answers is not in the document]
diegoferigo: Yes, all of them. Changed.

toolbox/include/base/Configuration.h
```diff
@@ -0,0 +1,162 @@
...
+
+    // GET METHODS
+    // ===========
+
+    /**
+     * Set the name of the robot
+     *
+     * @return Name of the robot
+     */
+    std::string getRobotName() const;
```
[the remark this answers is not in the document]
diegoferigo: Yes, all of them. Changed.

toolbox/include/base/Log.h
```diff
@@ -0,0 +1,42 @@
...
+    std::vector<std::string> errors;
+    std::vector<std::string> warnings;
+    std::string prefix;
+
+    static std::string serializeVectorString(std::vector<std::string> v, std::string prefix="");
+public:
+
+    static wbt::Log& getSingleton();
+
+    void Error(std::string errorMessage);
```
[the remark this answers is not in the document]
diegoferigo: In this case I liked visually the capital letter ^^ Changing.

toolbox/include/base/RobotInterface.h
```diff
@@ -0,0 +1,281 @@
...
+ *         supported. If in the future the support of operating on a subset will be implemented,
+ *         IPositionControl2 and IVelocityControl2 should be implemented. For the time being,
+ *         a possible workaround for this situation is creating a new configuration block
+ *         containing only the reduced set in a deeper-hierarchy Simulink's subsystem.
+ * @see RobotInterface::getDevice
+ */
+struct wbt::YarpDevices
+{
+    std::shared_ptr<yarp::dev::IControlMode2> iControlMode;
+    std::shared_ptr<yarp::dev::IPositionControl> iPos;
```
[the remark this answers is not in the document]
diegoferigo: Not at all

toolbox/include/base/RobotInterface.h
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     * Converts a vector of strings into a string format \code (element1 element2 element3) \endcode.
+     * It optioanlly support appending a prefix to the elements. e.g.
+     * \code ({prefix}element1 {prefix}element2 {prefix}element3)
+     *
+     * @param  v      The vector of strings
+     * @param  prefix The optional element prefix
+     * @return        The serialized string
+     */
+    static std::string vector2string(std::vector<std::string> v, std::string prefix="");
```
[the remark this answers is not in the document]
diegoferigo: :+1:

toolbox/include/base/RobotInterface.h
```diff
@@ -0,0 +1,281 @@
...
+     * @return A \c shared_ptr to the KinDynComputations object
+     */
+    const std::shared_ptr<iDynTree::KinDynComputations> getKinDynComputations();
+
+    /**
+     * Get the IControlMode2 interface of the robot
+     *
+     * @return a \weak_ptr of the interface
+     */
+    const std::weak_ptr<yarp::dev::IControlMode2> getIControlMode2();
```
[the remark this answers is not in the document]
diegoferigo: Changed as discussed f2f

toolbox/include/base/RobotInterface.h
```diff
@@ -0,0 +1,281 @@
...
+        // Blocks which require the RemoteControlBoardRemapper need to retain / release it
+        // in their initialization / terminate phase;
+        assert(m_robotDevice);
+        if (!m_robotDevice) {
+            // Return an empty weak pointer
+            return std::weak_ptr<T>();
+        }
+
+        T* ptr = device.get();
+        if (!m_robotDevice->view(ptr)) {
```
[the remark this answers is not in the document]
diegoferigo: It is not, calling the `reset()` method is needed

toolbox/include/base/ToolboxSingleton.h
diegoferigo: Used `= delete`

toolbox/include/base/ToolboxSingleton.h
diegoferigo: This was an inheritance of the previous version, changed the name. This method might even be removed, it's never called. Right now the best way to gather the model is asking it directly from the parent `WBBlock` using `getRobotInterface()->getKinDynComputations()`. This allows the block not to know the existence of the `confKey` variable.

toolbox/include/base/WBBlock.h
diegoferigo: Removed pointers and switched to class objects·

toolbox/include/base/ToolboxSingleton.h
diegoferigo: Renamed to `storeConfiguration()`

toolbox/include/base/WBBlock.h
diegoferigo: This is the key of the map contained in `ToolboxSingleton`, `ID` seems too generic in my opinion. The block itself should not be aware of this key, and gathering the correct information is handled directly from its `WBBlock` parent

toolbox/src/base/RobotInterface.cpp
```diff
@@ -0,0 +1,258 @@
...
+    if (m_robotDeviceCounter > 1) {
+        m_robotDeviceCounter--;
+        return true;
+    }
+
+    // This should be executed by the last block which uses CB (m_robotDeviceCounter=1)
+    assert(m_robotDevice);
+    if (m_robotDevice) {
+        //  Close the device
+        m_robotDevice->close();
```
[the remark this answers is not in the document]
diegoferigo: :+1:

toolbox/src/base/RobotInterface.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+// INITIALIZATION HELPERS
+// ======================
+
+bool RobotInterface::initializeModel()
+{
+    assert (!m_kinDynComp);
+
+    // Allocate the object
+    m_kinDynComp = std::make_shared<iDynTree::KinDynComputations>();
```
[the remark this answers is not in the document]
diegoferigo: I found this https://stackoverflow.com/questions/20895648/difference-in-make-shared-and-normal-shared-ptr-in-c

toolbox/src/base/Signal.cpp
diegoferigo: I quickly changed `<type> *var` to `<type>* var`, this is a mistake :sweat_smile: Anyway, I'll pass everything through a `clang-format` at the very end. Now it would make commits too dirty and difficult to squash / reorder

toolbox/src/base/Signal.cpp
diegoferigo: As above

toolbox/src/base/Signal.cpp
diegoferigo: As discussed f2f, there is currently a bug of the memory management of all signal types excluding `double`. I will refactor `Signal` and `Data` classes with templates and std objects in a second stage.

toolbox/src/base/Signal.cpp
diegoferigo: Replied [here](https://github.com/robotology/WB-Toolbox/pull/56#discussion_r150213326)

toolbox/src/base/ToolboxSingleton.cpp
diegoferigo: The conversion from shared_ptr to bool is `explicit`. An alternative to `get()` would be `static_cast<bool>`. Using this latter version.

toolbox/src/base/WBBlock.cpp
diegoferigo: Replied above

toolbox/src/base/WBBlock.cpp
diegoferigo: I agree, it is cleaner and less error prone. This must go in the new developers documentation.

toolbox/src/base/RobotInterface.cpp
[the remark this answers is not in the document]
francesco-romano: Great!
I am a bit scared of
```
Since there there's only one allocation, the pointee's memory cannot be deallocated until the control block is no longer in use. A weak_ptr can keep the control block alive indefinitely.
```
but I think I have to improve a bit my knowledge of shared ptrs before understanding correctly this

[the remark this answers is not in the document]
diegoferigo: By the end of the story, it's just the container that stores the counter. In any case its memory footprint would be almost 0.

toolbox/include/base/RobotInterface.h
```diff
@@ -0,0 +1,281 @@
...
+ * with the specified robot (real or model).
+ *
+ * @see wbt::Configuration
+ * @see wbt::YarpDevices
+ * @see iDynTree::KinDynComputations
+ */
+class wbt::RobotInterface
+{
+private:
+    std::shared_ptr<yarp::dev::PolyDriver> m_robotDevice;
```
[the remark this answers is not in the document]
diegoferigo: Moving the templated `getDevice()` function to the cpp, yes. It is possible since it is a private method.

toolbox/include/base/WBBlock.h
diegoferigo: :+1:

MxAnyType/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+project(MxAnyType LANGUAGES CXX VERSION 0.1)
+cmake_minimum_required(VERSION 3.0.2)
+
+# Configure the project
+# =====================
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+include(GNUInstallDirs)
+option(BUILD_SHARED_LIBS OFF)
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
traversaro: This variable only make sense if you are building shared libraries.

MxAnyType/CMakeLists.txt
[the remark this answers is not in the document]
traversaro: What do you mean by "been fixed"? And how this is related to PRIVATE/PUBLIC?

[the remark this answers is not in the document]
francesco-romano: @traversaro the actual question is: Eigen members in header files.
I know there were issues with exported symbols in Windows. Furthermore, I do not like them in header files (as they force you to have Eigen)

MxAnyType/CMakeLists.txt
diegoferigo: Yes, in the future this will become a shared library. For the time being it is static, but just con convenience.

toolbox/autogenerated/CMakeLists.txt
traversaro: The `project` CMake command in the end is not really something special, it just defines a bunch of variables. However, I see two possible problems here: 
* It is not clear what it is the version declared here, 
* A non-generated file in a directory `autogenerated` ? I typically try to keep only auto-generated files in a directory called `autogenerated`, but perhaps it is just a personal taste.

MxAnyType/include/AnyType.h
```diff
@@ -0,0 +1,65 @@
+#ifndef ANYTYPE_H
+#define ANYTYPE_H
+
+#include <Eigen/Core>
```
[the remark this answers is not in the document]
traversaro: I don't see any problem at all if `AnyType.h` is not part of a public library, i.e. a C++ library that we will expect to use outside the WB-Toolbox repository. Even if `AnyType.h` will eventually be exported as a C++ library, including the Eigen headers in the public headers is not a impossible, but it creates the following problems: 
* it complicates writing the `PackageConfig.cmake` file,
* it can create problems if there are several Eigen version in the system/bazel workspace/cmake superbuild (unfortunately some projects vendor their own non-system Eigen version, such as TensorFlow or Drake),
* it can create ABI incompatibilities if downstream user/libraries change the Eigen ABI using the Eigen compile-time options (see https://github.com/flexible-collision-library/fcl/issues/96#issuecomment-205229439).

toolbox/src/base/RobotInterface.cpp
```diff
@@ -0,0 +1,447 @@
...
+            return std::weak_ptr<T>();
+        }
+
+        T* ptr = nullptr;
+        if (!m_robotDevice->view(ptr)) {
+            // Return an empty weak_ptr
+            return std::weak_ptr<T>();
+        }
+        // Store ptr into the smart pointer
+        device.reset(ptr);
```
traversaro: This line is transferring ownership of the object pointer by `ptr` to the shared_ptr, but the ownership of the object (i.e. who will destroy it) is still maintained by the PolyDriver object, and there is no way to share it, and it is not implemented using smart pointers. I am surprised that this is not causing any kind of "double free" problems.

traversaro: @diegoferigo @francesco-romano

MxAnyType/CMakeLists.txt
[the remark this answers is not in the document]
traversaro: Replied in https://github.com/robotology/WB-Toolbox/pull/56#discussion_r150419203 .

toolbox/include/base/Configuration.h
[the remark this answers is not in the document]
traversaro: Why is the `std::array:data()` http://en.cppreference.com/w/cpp/container/array/data  not sufficient for this?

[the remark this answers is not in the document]
diegoferigo: It is what I was looking for, and I didn't at the time for some reason :sweat_smile:

toolbox/src/base/RobotInterface.cpp
diegoferigo: @traversaro I was not aware of this, and maybe I did not completely got your point here. This (templated) method returns a weak pointer, and it should not deallocate the object in its destructor, am I wrong?

toolbox/autogenerated/CMakeLists.txt
diegoferigo: This folder contains autogenerated code, but is copied manually here. I already discussed last week with @francesco-romano your same concern about the name of this folder. What we would need is the installed thrift file, that currently nor yarp and gazebo-yarp-plugins install. We agreed on waiting one of those projects (yarp if I recall well) to handle this installation in order to clean this folder and let it become a real autogenerated directory.

For what concerns the CMake file, with this logic is indeed useless having a separate project , it will become a target as soon as this autogeneration will work as expected. For the time being if you don't have any concern I would leave this as it is and open an issue for future reference.

toolbox/src/base/RobotInterface.cpp
traversaro: > A solution I can see is using weak_ptr also for the internal members. Thanks a lot for this comment, I was not aware of the implementation of the device.

For having a `weak_ptr` a shared_ptr must have the ownership of the object, and this is not the case. The only solution I can think of is using raw pointers.

diegoferigo: Yes, no other clean solution exist unfortunately. We'll live well anyways :smile:

MxAnyType/include/AnyType.h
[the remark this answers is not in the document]
diegoferigo: Thanks @traversaro! Let's keep things simple and let's use only `std::vector` for returning data. The only issue I see here is efficiency, the raw array created by Matlab is copied once.

Actually, it is copied twice, because mex functions already perform a deep copy of the internal representation used by Matlab. There are however undocumented functions which allow avoiding this copy, and I'd like to check them sooner or later. It would be nice having in the interface also the support of 0-copy methods (for sure we would have to return C-style arrays), handy when data is big. This is not our case, and this will be a future activity.

MxAnyType/CMakeLists.txt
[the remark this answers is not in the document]
diegoferigo: Final solution https://github.com/robotology/WB-Toolbox/pull/56#discussion_r150479807

MxAnyType/include/AnyType.h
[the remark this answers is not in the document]
traversaro: If this code is only used at configuration time, I would worry for efficiency only if it emerges as a problem in actual usage. Having said that, I think the cleanest way to support this kind of pattern is to implement a simple structure;
~~~cpp
struct ExternalVector
{
    ExternalVector() {}
    ExternalVector(double * in_data, size_t in_size): data(in_data), size(in_size) {} 
    double * data{nullptr};
    size_t size{0};
}
~~~
and add a `asVectorDouble(const ExternalVector & data)` method.  
In this way the user-code can provide a simple function: 
~~~cpp
inline ExternalData fromEigen(Eigen::VectorXd& vec) { return ExternalData(vec.data(), vec.size()); }  
~~~
And happy use the method without any additional copy.

diegoferigo:
@francesco-romano @traversaro The toolbox development is almost done (see bullet list in the first post). C++ code is now ready to be tested block after block. Let me know if you have any other concern about the current state of the code.

toolbox/include/CentroidalMomentum.h
```diff
@@ -1,35 +1,39 @@
...
+namespace iDynTree {
+    class SpatialMomentum;
+}
 
-    double *m_basePose;
-    double *m_centroidalMomentum;
+class wbt::CentroidalMomentum : public wbt::WBBlock
+{
+private:
+    std::unique_ptr<iDynTree::SpatialMomentum> m_centroidalMomentum;
```
traversaro: There is probably a reason, but I don't remember it: why it is not possible to declare this variable simply as class members?

traversaro:
@diegoferigo What is the plan regarding https://github.com/robotology/WB-Toolbox/pull/56/files#r150419463 ?

toolbox/include/CentroidalMomentum.h
diegoferigo: Headers cleanliness I'd say. This was the pattern used in the previous release, and I kept it also in this new version.

diegoferigo:
@traversaro As discussed above. I push the fix as soon as I test the first block which uses yarp interfaces.

diegoferigo:
@francesco-romano @traversaro 
With the latest commits the toolbox is 99% ready. Only minor blocks still need to be tested, most of them work as expected when used independently.

In the next few days I'll  start the test of more complex controllers.

p.s. @traversaro for the fix of the interfaces, look b62e3925146a4fc85341b368f4785c6e64a881bc

diegoferigo:
The `WB-Toolbox 3.0` is finally ready @francesco-romano @traversaro @DanielePucci.

All the blocks have been tested and seem working. The simple `impedanceControl` seems working fine. As soon as @nunoguedelha concludes the work on the standing positioning of the new `iCubGazeboV2_5` I'll test the YOGA demo.

francesco-romano:
Great work @diegoferigo 
As soon as I have some time, I will check the rest of the PR.

toolbox/include/CentroidalMomentum.h
```diff
@@ -1,35 +1,36 @@
...
+{
+private:
+    iDynTree::SpatialMomentum* m_centroidalMomentum;
 
-    //input buffers
-    double *m_basePoseRaw;
-    double *m_configuration;
-    double *m_baseVelocity;
-    double *m_jointsVelocity;
+    static const unsigned INPUT_IDX_BASE_POSE;
```
francesco-romano: Just a couple of comments on this:
- If you want to adopt the CamelCase style instead of C-Style Macro style for me is ok
- what are the pro/cons of private static const variables vs static global variables in the Cpp file?

toolbox/src/CentroidalMomentum.cpp
```diff
@@ -2,158 +2,167 @@
...
+    bool CentroidalMomentum::terminate(BlockInformation* blockInfo)
     {
-        if (m_basePose) {
-            delete [] m_basePose;
-            m_basePose = 0;
-        }
         if (m_centroidalMomentum) {
-            delete [] m_centroidalMomentum;
+            delete m_centroidalMomentum;
             m_centroidalMomentum = 0;
```
francesco-romano: for consistency use `nullptr`

toolbox/src/DotJNu.cpp
```diff
@@ -1,186 +1,216 @@
...
     {
-        if (m_basePose) {
-            delete [] m_basePose;
-            m_basePose = 0;
-        }
+        // Output
         if (m_dotJNu) {
-            delete [] m_dotJNu;
+            delete m_dotJNu;
             m_dotJNu = 0;
```
francesco-romano: same comments as `CentroidalMomentum`

toolbox/include/DotJNu.h
```diff
@@ -1,38 +1,40 @@
...
+    iDynTree::FrameIndex m_frameIndex;
 
-    //input buffers
-    double *m_basePoseRaw;
-    double *m_configuration;
-    double *m_baseVelocity;
-    double *m_jointsVelocity;
-
-    int m_frameIndex;
+    static const unsigned INPUT_IDX_BASE_POSE;
```
francesco-romano: same comments as `CentroidalMomentum`

toolbox/include/CentroidalMomentum.h
diegoferigo: > If you want to adopt the CamelCase style instead of C-Style Macro style for me is ok

I add this as Post-WB3.0 task, I have nothing against it.

> What are the pro/cons of private static const variables vs static global variables in the Cpp file?

No big differences, actually moving them in the cpp keeps headers cleaner.

PR #77, opened 2018-02-22
[WIP] Automatic Code Generation Support

diegoferigo:


diegoferigo:
This is a huge PR @traversaro. Understanding the logic behind the automatic code generation is a bit tricky and if you are interested in details we can have a f2f meeting (waiting for some detailed documentation and tutorials on the website).

For the time being, it would be nice receiving feedback mainly on:

- Templates and explicit specialization & declaration: [`ConvertStdVector.h`](https://github.com/robotology/wb-toolbox/pull/77/files#diff-f96e3e95a96c78a0cb8afa3b30c5f33c) [`ConvertStdVector.cpp`](https://github.com/robotology/wb-toolbox/pull/77/files#diff-d05ba6c0ae809a5f0fc5be4f24e8dbe4) [`Parameters.h`](https://github.com/robotology/wb-toolbox/pull/77/files#diff-e69302e9f60aec0c5e5401c9c1061d48) [`Parameters.cpp`](https://github.com/robotology/wb-toolbox/pull/77/files#diff-1c6221c8d3e5e2cd6e30274e3da2bbe8)
- Pimpl idiom is edge cases: [`Configuration.cpp`](https://github.com/robotology/wb-toolbox/pull/77/files#diff-3ea2a5a8c74ff3831297147eac1ba95e)

toolbox/CMakeLists.txt
```diff
@@ -202,57 +221,136 @@ configure_block(BLOCK_NAME "Get Limits"
...
+    ${ALL_HEADERS} ${ALL_SOURCES}
+)
+set_target_properties(WBToolboxLibrary PROPERTIES
+    OUTPUT_NAME "WBToolbox"
+    # VERSION ${PROJECT_VERSION}
+    PUBLIC_HEADER "${ALL_HEADERS}"
+)
+
+# Import math symbols from standard cmath
+target_compile_definitions(WBToolboxLibrary PUBLIC "-D_USE_MATH_DEFINES")
```
traversaro: Why `PUBLIC` ? I far as I can see, `M_PI` is used only in `.cpp` files. I would turn it `PRIVATE` and only add it only when compiling on `MSVC` ?

toolbox/CMakeLists.txt
```diff
@@ -202,57 +221,136 @@ configure_block(BLOCK_NAME "Get Limits"
...
-include_directories(SYSTEM ${YARP_INCLUDE_DIRS} ${EIGEN3_INCLUDE_DIR})
+target_include_directories(WBToolboxLibrary PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/base>
+)
+
+target_include_directories(WBToolboxLibrary SYSTEM PUBLIC
+    ${Matlab_INCLUDE_DIRS}
+    ${Matlab_ROOT_DIR}/simulink/include
+    ${YARP_INCLUDE_DIRS}
```
traversaro: I think this is not necessary, and this should be supported just by linking YARP .

toolbox/include/base/Factory.h
```diff
@@ -1,31 +1,37 @@
...
 #include "InverseDynamics.h"
 #include "Jacobian.h"
 #include "MassMatrix.h"
-#include "ModelPartitioner.h"
-#include "RealTimeSynchronizer.h"
-#include "SetLowLevelPID.h"
-#include "SetReferences.h"
-#include "SimulatorSynchronizer.h"
 
+// iCub-depent blocks
```
traversaro: typo: dependent

toolbox/include/base/RobotInterface.h
```diff
@@ -32,225 +35,118 @@ namespace iDynTree {
...
-    typedef std::unordered_map<cb_idx, max_cb_idx> ControlBoardIdxLimit;
+    class Configuration;
+
+    using JointIndex_Yarp = int;
+    using JointIndex_iDynTree = int;
+    using JointName = std::string;
+    using ControlBoardIndex = unsigned;
+    using JointIndexInControlBoard = unsigned;
+
+    using JointIndexToYarpMap =
```
traversaro: I am a bit confused by all this structures (and avoiding this is actually why the remapper and the support for specifying the serialization in iDynTree was explicitly coded to avoid this constructs!). 
If this is just used for the "smandruppator", in the long run I suggests to rename this variables to be a generic "subpart" division of the joints, as in the future we could easily connect to the robot using a single controlboard containing all the joints (for performance reason) while keeping the nice division of the robot in parts for visualization/debugging.

toolbox/CMakeLists.txt
```diff
@@ -99,8 +118,6 @@ if (WBT_USES_ICUB)
     #
     #     include_directories(SYSTEM ${iKin_INCLUDE_DIRS})
     # endif()
-
-    include_directories(SYSTEM ${ctrlLib_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
diegoferigo: This didn't disapper, it was just moved after the creation of the `WBToolboxLibrary` target. Here's the code (every reference might disappear due to a force push):

```cmake
if (WBT_USES_ICUB)
    list(APPEND LINKED_LIBRARIES ctrlLib)
    target_compile_definitions(WBToolboxLibrary PUBLIC -DWBT_USES_ICUB)
    target_link_libraries(WBToolboxLibrary PUBLIC ctrlLib)
    if (${ICUB_USE_IPOPT})
        list(APPEND LINKED_LIBRARIES iKin)
        target_link_libraries(WBToolboxLibrary PUBLIC iKin)
    endif()
endif()
```

I switched to `target_*` when possible, and includes are included transitively.

toolbox/CMakeLists.txt
diegoferigo: Good catch!

toolbox/CMakeLists.txt
diegoferigo: Yes they are not required. Linking YARP is sufficient. By the way, would you recommend to limit the YARP components as I do with iDynTree?

toolbox/include/base/Factory.h
diegoferigo: Thanks!

toolbox/include/base/RobotInterface.h
diegoferigo: All these maps and method for gathering them are for the "smandruppator" (ModelPartitioner). I plan sooner or later to polish this RobotInterface class further and use it as standalone wrapper for the RemoteControlBoardRemapper and KinDynComputations. Since these maps might be helpful and coding them was a real pain, I decided to store them here. They are lazily accessed, if you never ask the get methods the code that spawns multiple single control board is never executed.

toolbox/CMakeLists.txt
```diff
@@ -10,21 +13,31 @@ if(NOT ${YARP_IS_SHARED_LIBRARY})
     message(FATAL_ERROR "YARP was found, but it was compiled as static library. A shared library version of YARP is required.")
 endif()
 
+find_package(Matlab REQUIRED
+    MX_LIBRARY
+    ENG_LIBRARY
+    MAIN_PROGRAM
```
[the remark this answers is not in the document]
traversaro: Ahhh, I got it.

toolbox/CMakeLists.txt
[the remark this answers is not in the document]
traversaro: >  includes are included transitively.

Unfortunately, this does not hold for ICUB libraries, as  they do not declare they include libraries using `target_include_libraries`, see https://github.com/robotology/icub-main/search?utf8=%E2%9C%93&q=target_include_directories&type= . For this reason, using directly `ctrlLib_INCLUDE_DIRS` and `iKin_INCLUDE_DIRS` is necessary.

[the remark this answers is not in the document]
diegoferigo: I see -.- This is unfortunately misleading. I usually install all icub stuff in the same folders as yarp and I would never notice it. Fixed.

toolbox/src/base/ConvertStdVector.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+template <>
+void wbt::convertStdVector<int, std::string>(const std::vector<int>& input,
+                                             std::vector<std::string>& output)
+{
+    output.clear();
+    output.resize(input.size());
+    std::transform(input.begin(), input.end(), output.begin(), [](const int& num) {
+        return std::to_string(num);
```
traversaro: How are you using this methods? If you are using this methods to convert parameters in configuration files, typically you do not want this conversion to be influenced by the locale setting, and so `to_string` is not the appropriate functions. I suspect also `std::stod` is affected. 

I originally became aware of this edge case in the context of URDF parsers, see https://github.com/robotology/idyntree/issues/288 . Most of the time this kind of problems it not recognized because:
* The default locale setting for C/C++ program is `"C"`. 
* The user-preferred language for most developers is `en-US`, or in general `en-SOMETHING`. 

However, if the libraries is used in a process that uses the Qt libraries on a computer in which the user-preferred locale is something like `it-IT` or `es-ES`, the Qt libraries set the C/C++ locale to be the user-preferred locale, and the libraries stop working properly. As you may imagine, this is not a behavior that is easy to debug.

toolbox/include/base/Parameters.h
```diff
@@ -0,0 +1,265 @@
...
+ */
+class wbt::Parameters
+{
+public:
+    using ParamIndex = int;
+    using ParamName = std::string;
+
+private:
+    class impl;
+    std::unique_ptr<impl, void (*)(impl*)> pImpl;
```
traversaro: I am confused here. You are passing `void (*)(impl*)` as the deleter, and I am not sure that I understand what this mean at the C++ level to be honest. Who is actually deleting the class? Are you sure that this is not creating a memory leak?

traversaro: Sorry, I got it during the review. `void (*)(impl*)` is the type of the deleter, and the actual deleter is passed in the unique_ptr constructor. I think you mentioned me f2f that the need for doing this is due to the necessity of defining a copy constructor, it would be nice to have a small commenter here hinting that.

toolbox/include/base/Parameters.h
```diff
@@ -0,0 +1,265 @@
...
+     * @brief Store a scalar parameter
+     *
+     * @tparam The type of the parameter to store. Despite this, the parameter get cast accordingly
+     *         to its metadata.
+     * @param param The value of the scalar parameter to store.
+     * @param paramMetadata The metadata associated to the parameter to store.
+     * @return True for success, false otherwise.
+     */
+    template <typename T>
+    bool storeParameter(const T& param, const wbt::ParameterMetadata& paramMetadata);
```
traversaro: Have you tried compiling and linking this on Windows? I think the `WINDOWS_EXPORT_ALL_SYMBOLS` option will be quite stressed by this edge case.

traversaro: Also giving it a shot on macOS may be worth, given all the intensive use of templates.
You can use my former iMac (the idea is using it for that). cc @DanielePucci

toolbox/src/base/ConvertStdVector.cpp
diegoferigo: > As you may imagine, this is not a behavior that is easy to debug.

I remember "piovere madonne" from that issue :sweat_smile: In this case no file is involved, at least, not at the current stage.

This template is used inside `wbt::Parameters` for converting vector parameters between different data types. Values are usually read from the block's mask in Simulink.

I don't see how this can affect `std::to_string`, but I see the point of the other way around, and I mean `std::stod` and `std::stoi` which I use in many points. The origin of this feature is to give the possibility of calling:

```cpp
// Store a parameter as a double
Parameter p(42, {ParameterType::DOUBLE, 0, 1, 1, "ParamName"});

// Insert it in a container with other parameters
Parameters params;
params.storeParameter(p);
...

// Get the parameter that is cast internally
int value;
getParameter("ParamName", value);
```

This should hold also for `std::string`.

In future versions I plan to give the possibility of override those variables for give flexibility to the automatic code generation. In this case, probably the parameters to override will be stored in a configuration file.

@traversaro Do you think this problem should be addressed by whom will parse the configuration? It will not be responsibility of the toolbox. And, if this is the case, all `RFModule` should be careful of this behavior (how does yarp handle it? I guess it does not).

traversaro: > And, if this is the case, all RFModule should be careful of this behavior (how does yarp handle it? I guess it does not).

I do not think that `yarp::os::Property` use any kind of double to string facilities provided by C++. 

> I don't see how this can affect std::to_string, but I see the point of the other way around

Basically, you risk to store doubles as `10,000`. Depending on the usage, it may or not may be a problem. 

> Values are usually read from the block's mask in Simulink.

They are read as strings or directly as doubles? In the generated code, how they are stored? If they are read or stored as strings, I think we already have the problem, as a user would but `0.10` in its Simulink mask, and in could be read if the wrong way depending on the locale.

toolbox/include/base/Parameters.h
diegoferigo: Yes something like that. If I understood correctly, the issue here is use usage of:

```cpp
~WBBlock() override = default;
```

I used as an example `WBBlock` which is **not** copiable.

The combination of the default destructor plus the forward declaration of the object handled by the `unique_pointer` makes `sizeof` complain. `sizeof` is used in the default deleter, for this reason passing a custom deleter solves this issue.

toolbox/include/base/Parameters.h
diegoferigo: > Have you tried compiling and linking this on Windows?

Of course no :grin: We actually claim multiplatform support but we never tested broadly on non-linux platforms. For this reason some while ago I decided to put a [big disclaimer](https://robotology.github.io/wb-toolbox/mkdocs/install/#install) on the website.

toolbox/include/base/Parameters.h
traversaro: > The combination of the default destructor plus the forward declaration of the object handled by the unique_pointer makes sizeof complain. sizeof is used in the default deleter, for this reason passing a custom deleter solves this issue.

The combination of the default destructor plus the forward declaration of the object handled by the unique_pointer is a standard feature of the pimpl pattern, see for example all the usage in the blocks such as https://github.com/robotology/wb-toolbox/pull/77/files#diff-d7d137bec79209b2e9a3a4548fd95e51R25, and there seems not to be problem with `sizeof` .

If I understood correctly, the problem is combining this with a non-copyable class?  In that case it is not clear to me how it is related to sizeof, but I do not strongly feel about this issue, so if it is working for you it is fine for me.

toolbox/include/base/Parameters.h
traversaro: No, apparently we cannot use the iMac because it does not have a Matlab/Simulink license.

toolbox/src/base/ConvertStdVector.cpp
diegoferigo: > They are read as strings or directly as doubles?

Numeric types are read as `double` and strings as `string`s. Simulink functions are used, and we do not have control on them. Hopefully they are robust to locale.

> In the generated code, how they are stored? 

About serialization / deserialization to the RTW file (which is kind of a database read by the automatic generation pipeline) the following logic is used:

- Strings are serialized as `"stringContent"`
- Numeric types are all converted to `double` and serialized as `42.0`

Again, I use Simulink functions for doing this.

If what I wrote so far is right, the parameters are correctly read by Simulink, correctly serialized to the RTW file, and correctly read from the RTW file from the autogenerated code pipeline.

### Possible problems I see

These helpers are used only when the block developer wants for some reason to get a parameter with a type different than the original type (specified in the parameter metadata). This step might be locale-dependent.

#### Numeric -> string

If you want to convert a double to a string you will get `"10,000"` as you mentioned. But, since this is already kind of downstream, after the cast unlikely you will compare numeric types as strings. Rather than converting the numeric parameter to string you would rather convert the other string to numeric and then compare the numeric values.

#### String -> numeric

Problems would arise if the block developer chooses to store a numeric type as a string (non-default behavior). In this case I have no idea what would be the result of calling `std::stod("42.0")` on locales that assume the `"42,0"` as a separator. Anyway, this is a very edge case.

### Final thoughts

In the light of these considerations, the current situation doesn't look problematic. Does it sound good even for you @traversaro?

toolbox/include/base/Parameters.h
diegoferigo: From [unique_ptr](http://en.cppreference.com/w/cpp/memory/unique_ptr):

> std::unique_ptr may be constructed for an incomplete type T, such as to facilitate the use as a handle in the pImpl idiom. If the default deleter is used, T must be complete at the point in code where the deleter is invoked, which happens in the destructor, move assignment operator, and reset member function of std::unique_ptr.

As discusses f2f, when possible removing the custom deleter and moving the default destructor to the cpp is preferred for readability.

toolbox/src/base/CoderBlockInformation.cpp
```diff
@@ -0,0 +1,327 @@
...
+    if (pImpl->parametersFromRTW.getNumberOfParameters() == 0) {
+        wbtError << "The Parameters object containing the parameters to parse is empty.";
+        return false;
+    }
+
+    for (wbt::ParameterMetadata md : pImpl->paramsMetadata) {
+        // Check that all the parameters that are parsed have already been stored from the coder
+        if (!pImpl->parametersFromRTW.existName(md.name)) {
+            wbtError << "Trying to get a parameter value for " << md.name
+                     << ", but its value has never been sored.";
```
traversaro: sored --> stored

toolbox/src/base/ConvertStdVector.cpp
traversaro: Sorry, I think I finally got it. 

The `Parameters` class is a generic key-value storage in which values are typed (either `std::string`, `bool`, `int` or `double`) used to set parameters for code-generated blocks. 
However, even if the parameters are typed, the class offers a way of converting internally parameters from one type to another. However, this feature is not used currently in any part of the code,  and at the moment we are not currently plan to use it, right? 

At this point, the only thing I am afraid is the current situation:
* a given block takes in input a `double` parameter, but for some reason the external C++ code save it as a string parameter, with some strange locale setting
* if I understand correctly the code in https://github.com/robotology/wb-toolbox/pull/77/files/05bd751d6d557cb110d1437bb845ba4a7812fcd0#diff-169c37a45eec0ad411c3e3a6f9052cf3R208, the parameter parsing is successful, but the parameter is actually converted implicitly from string to double. 

I must say that if I got the big picture, I do not understand the need for the implicit conversion functionality in `Parameters`, but now that it is there I do not think it make sense to try to remove it. 

I have nevertheless two suggestions, but I am not sure if they make sense:
* Add a line in `Parameters`'s documentation to say that for implicit conversions between std::string and the other types the standard functions `std::to_string`, `std::stoi`, etc etc are used, and the behavior of this functions are influenced by the process-global behavior.
* If it is not too complicated, consider also the implicit type when performing the check on the parameters.

toolbox/include/base/Parameters.h
```diff
@@ -0,0 +1,265 @@
...
+ * @brief Class for storing block's parameters
+ *
+ * This class can contain scalar and vector parameters of the supported types.
+ *
+ * @see wbt::Parameter, wbt::ParameterMetadata, wbt::ParameterType
+ */
+class wbt::Parameters
+{
+public:
+    using ParamIndex = int;
```
traversaro: For indexing purposes, if you need an unsigned quantity I would go for `std::ptrdiff_t`, see http://eigen.tuxfamily.org/index.php?title=3.3#Index_typedef .

toolbox/include/base/Parameters.h
```diff
@@ -0,0 +1,265 @@
...
+     * @param paramMetadata The metadata associated to the parameter to store.
+     * @return True for success, false otherwise.
+     */
+    template <typename T>
+    bool storeParameter(const T& param, const wbt::ParameterMetadata& paramMetadata);
+
+    /**
+     * @brief Store a vector parameter
+     *
+     * @tparam The type of the parameter to store. Despite this, the parameter get cast accordingly
```
traversaro: This comment also applies to the other public template methods: if I understood correctly, the template type can only be `bool`, `int`, `double` or `std::string`. Do you think it make sense to document that the type of the template can be only one of these four types?

MxAnyType/CMakeLists.txt
```diff
@@ -10,42 +15,56 @@ include(GNUInstallDirs)
...
-    set_target_properties(${VARS_PREFIX} PROPERTIES COMPILE_FLAGS "-fPIC")
-endif()
+set_target_properties(${VARS_PREFIX} PROPERTIES
+        VERSION ${PROJECT_VERSION} # This is for the symlink of the .so
+        PUBLIC_HEADER "include/AnyType.h;include/MxAnyType.h"
+)
 
 # Find Matlab resources
 find_package(Matlab REQUIRED MX_LIBRARY)
-target_include_directories(${VARS_PREFIX} SYSTEM PUBLIC "${Matlab_INCLUDE_DIRS}")
```
traversaro: Why did you remove the `"` in `"${Matlab_INCLUDE_DIRS}"` . `"` are typically used when the variable contains directory names with spaces.

toolbox/src/base/ConvertStdVector.cpp
diegoferigo: > However, this feature is not used currently in any part of the code, and at the moment we are not currently plan to use it, right?

We do not currently use it, but on the long run I thought it was a nice addition especially if we split the core of the toolbox. Consider the following snippet:

```cpp
Parameters parameters;
parameters.storeParameter(42, {ParameterType::DOUBLE,
                               /*index=*/0,
                               /*rows=*/1,
                               /*cols=*/1,
                               "paramName"});
```

The type specified in the metadata **always** wins. The usage without conversion might have led to a template instantiation difficult to figure out. In this way we don't mind.

toolbox/include/base/Parameter.h
```diff
@@ -0,0 +1,128 @@
...
+    CELL_STRING,
+    // Struct
+    STRUCT_INT,
+    STRUCT_BOOL,
+    STRUCT_DOUBLE,
+    STRUCT_STRING,
+    STRUCT_CELL_INT,
+    STRUCT_CELL_BOOL,
+    STRUCT_CELL_DOUBLE,
+    STRUCT_CELL_STRING
```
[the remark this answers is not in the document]
diegoferigo: It is not straightforward indeed. If the parameter to parse is a Matlab's cell array or a struct, the values introduced by this enum greatly simplify the logic to differentiate all these cases.

 You can appreciate this reading the `SimulinkBlockInformation::parseParameters(wbt::Parameters& parameters)` function.

From the user point of view (and I mean the block developer, which might not be an expert on the toolbox back-end layer), parsing a heterogeneous list of block parameters containing e.g. a double, a string, and a struct with various fields translates in writing:

```cpp
bool MyBlock::parseParameters(BlockInformation* blockInfo)
{
    ParameterMetadata param1(ParameterType::DOUBLE, 0, 1, 1, "DoubleParam");
    ParameterMetadata param2(ParameterType::STRING, 1, 1, 1, "StringParam");
    // This is a single struct but the user has to specify the type of the field.
    // Note that the index is always 2.
    // The name of the parameter matches the name of the struct field.
    ParameterMetadata param3_1(ParameterType::STRUCT_STRING, 2, 1, 1, "NameOfStructField1");
    ParameterMetadata param3_2(ParameterType::STRUCT_DOUBLE, 2, 1, 1, "NameOfStructField2");
    ParameterMetadata param3_3(ParameterType::STRUCT_CELL_INT, 2, 1, 5, "NameOfStructField3");
    // This last field of the struct contains a cell array of 5 values

    bool ok = true;
    ok = ok && blockInfo->addParameterMetadata(param1);
    ok = ok && blockInfo->addParameterMetadata(param2);
    ok = ok && blockInfo->addParameterMetadata(param3_1);
    ok = ok && blockInfo->addParameterMetadata(param3_2);
    ok = ok && blockInfo->addParameterMetadata(param3_3);

    if (!ok) {
        wbtError << "Failed to store parameters metadata.";
        return false;
    }

    return blockInfo->parseParameters(m_parameters);
}
```

If you dig into the Matlab functions to parse all these variety of cases, you quickly realize it is quite difficult. This is the cleanest interface I could think to provide to the block developer.

As a side note, the reported example is not an edge case. We use something very similar for `WBBlock`.

toolbox/include/base/Parameters.h
diegoferigo: It surely does

MxAnyType/CMakeLists.txt
diegoferigo: Restored them :+1:

toolbox/include/base/Parameter.h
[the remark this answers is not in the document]
diegoferigo: When I find some time I'd like to provide some tutorials for toolbox users (e.g. creating a simple model and generating C++ from it), block developers (e.g. how to create a block), and toolbox developers (random tips and tricks).

A more thorough documentation on parameters will be reported there.

diegoferigo:
@traversaro Tests are now passing, if you don't have any other comment I merge soon this PR in devel. I tested the Yoga controller on Gazebo and all works fine.

toolbox/include/base/Parameter.h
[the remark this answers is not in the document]
traversaro: Ok!

traversaro:
> @traversaro Tests are now passing, if you don't have any other comment I merge soon this PR in devel. I tested the Yoga controller on Gazebo and all works fine.

Sorry, I lost this  mention due to all the commits. Let's merge!

Issue #94, opened 2018-02-27
RemoteControlBoardRemapper behaviour in edge usage cases


diegoferigo:
@S-Dafarra please fill with more information as discussed f2f

It might happen that the robot goes in HW fault or something unusual happens during the situation. In this case, the robot's control boards go offline, meanwhile the `WB-Toolbox`'s remote wrapper stays on and for some reason it does not get closed properly.

I assume this behavior is due to the current logic that handles the `RemoteControlBoardRemapper` resource which explanation goes out of the scope of this issue. Long story short, every block that needs to access the `RemoteControlBoardRemapper` increase a counter with `retainRemoteControlBoardRemapper()`, and its `terminate()` method decrease it with `releaseRemoteControlBoardRemapper()`. The first retain allocates the device, and the last release deallocates it.

This is due to the fact that blocks do not access directly the device (and a shared pointer in this case might do the counter job). Instead, blocks use the interfaces which are `view()`ed from it, and yarp in this case works with raw pointers.

Considering this limitation, it is not clear how to improve this logic. I list here some random ideas (one for the time being) which might need some more time to get structured better in my mind:

### 1. Move the ownership of the `PolyDriver` containing the `RemoteControlboardRemapper`

Some information to understand this idea:

**Current status:** The `PolyDriver` is contained in a `RobotInterface` object owned by the `ToolboxSingleton` as a shared pointer.

**Other important details:** With the new version (`3.1`), the `BlockInformation` implementation will be persistently stored in a `PWork`.

A possible idea would be to move its ownership to the `BlockInformation` of the first block that allocates it. A weak pointer generated by it could be stored instead in the `ToolboxSingleton` (to keep reference), and all the other blocks that need the same `RobotInterface` object can get a copy of the shared pointer. In this way, the `RobotInterface` (and the `PolyDriver` which is inside it) are deleted when the last block is destroyed, removing the need of this retain / release approach.

I like this idea also because its goes in the direction we are already taking, which is relying less and less on the singleton. Wrt to the previous toolbox versions, we already moved out most of the functionalities from it. As an example, in the next version the toolbox (`3.1`) the singleton is hidden inside the different `BlockInformation` implementations. `Blocks` are not aware of it. 

---

cc @gabrielenava @traversaro @francesco-romano

diegoferigo:
[...]

PR #115, opened 2018-05-21
New singleton logic

diegoferigo:
This is probably one of the biggest architecture change the wb-toolbox received after its refactoring from `WBI-Toolbox` 1.0.

The singleton among releases (2.0, 3.0, and the upcoming 4) always underwent changes, all towards a very defined direction. Its name changed accordingly.

A quick history overview:

- In the release 2.0 it was called `WBIInterface`, and it was handling the deprecated layer of the `whole-body-interface`. It was used by the old `WBIModelBlock` and `WBIBlock` classes. Model and the robot resources were managed internally by the yarp implementation of the `whole-body-interface`.
- The release 3.0 deprecated the `whole-body-interface` and substituted it directly with `iDynTree` calls and the interfacing with the robot was performed directly with the `RemoteControlBoardRemapper`. In this release a new `RobotInterface` class was introduced to store the `iDynTree` and `Yarp` resources, and a lot of logic previously contained in the `WBIInterface` singleton were moved in this new class. Since the interface deprecation, the singleton was renamed to `ToolboxSingleton`. As already mentioned, its functionalities shrunk, and it became a container with a singleton fashion of many `RobotInterface` objects (for the multi robot support). Due to this switch, some of the logic for gathering `iDynTree` and `Yarp` resources by the `Block`s required to temporary modify the `BlockInformation` interface adding calls to gather `KinDynComputations` and `RobotInterface` objects.
- The upcoming release 4 is a preparation release for splitting the core component of wb-toolbox, and this means that core classes contained in the `base` subfolders should be completely independent by these external dependencies. The singleton, which handle these resources, should not be part of this core component. This PR handled this last missing step.

In more detail, this PR introduces the following changes:

- `RobotInterface` and `KinDynComputations` are not gathered anymore from the `BlockInformation` interface. `WBBlock`s which need these resources ask them directly from the singleton. This didn't require a massive block refactoring because `WBBlock` already contained an helper method to gather these resources from the `BlockInformation` interface. Changing this method to ask them from the singleton was enough to implement this change.
- Since now only `WBBlock` instances need this singleton and these two entities are independent from the toolbox core, the singleton name changed again (hopefully for the last time) to `WholeBodySingleton`.
- **Important:** The singleton does not own anymore the memory of the `RobotInterface` object. Its ownership has been moved directly to the `WBBlock`s instances as a shared pointer (so many `WBBlock`s that share the same robot can use the same model and yarp interfaces), and the singleton only holds a weak pointer to the shared `RobotInterface`. This is required because the purpose of the singleton is to implement the logic for let this sharing happen, since single `WBBlock`s don't know anything about other `WBBlock`s.
- This change of ownership allowed to remove the hacky counter and retain / release logic for allocating and freeing the `RemoteControlBoardRemapper` device. Since it is contained by the `RobotInterface` object, which is shared among blocks, when the last block owning its memory get deleted, the `RobotInterface` is destructed, with its internal `RemoteControlBoardRemapper` object.
- As a consequence, this new logic greatly improves the robustness of the singleton, solving https://github.com/robotology/wb-toolbox/issues/94. It is more difficult now to reach a dirty state. Even if the `terminate` method is not called (which previously was deallocating with a counter the `RemoteControlBoardRemapper`), it is enough that the S-Function calls the `mdlTerminate` callback that destruct the objects in order to reach a clean state. And this is enforced by properly setting the Simulink engine.

Ok, pau for now. This was a long post but I though it was worth to store this information somewhere and this PR marks the end of a long refactoring path.

cc @traversaro @francesco-romano @DanielePucci

### DIFF
--- a/toolbox/src/InverseDynamics.cpp
+++ b/toolbox/src/InverseDynamics.cpp
@@ -78,12 +78,7 @@ bool InverseDynamics::configureSizeAndPorts(BlockInformation* blockInfo)
     }
 
     // Get the DoFs
-    const auto robotInterface = getRobotInterface(blockInfo).lock();
-    if (!robotInterface) {
-        wbtError << "RobotInterface has not been correctly initialized.";
-        return false;
-    }
-    const int dofs = robotInterface->getConfiguration().getNumberOfDoFs();
+    const int dofs = getRobotInterface()->getConfiguration().getNumberOfDoFs();
 
     // INPUTS
     // ======
@@ -139,12 +134,7 @@ bool InverseDynamics::initialize(BlockInformation* blockInfo)
     using namespace iDynTree;
 
     // Get the DoFs
-    const auto robotInterface = getRobotInterface(blockInfo).lock();
-    if (!robotInterface) {
-        wbtError << "RobotInterface has not been correctly initialized.";
-        return false;
-    }
-    const auto dofs = robotInterface->getConfiguration().getNumberOfDoFs();
+    const auto dofs = getRobotInterface()->getConfiguration().getNumberOfDoFs();
 
     // Initialize sizes and value
     pImpl->baseAcceleration.zero();
@@ -152,7 +142,7 @@ bool InverseDynamics::initialize(BlockInformation* blockInfo)
     pImpl->jointsAcceleration.zero();
 
     // Get the KinDynComputations pointer
-    const auto& kindyn = robotInterface->getKinDynComputations();
+    const auto& kindyn = getRobotInterface()->getKinDynComputations();
     if (!kindyn) {
         wbtError << "Failed to get the KinDynComputations object";
         return false;
@@ -175,7 +165,7 @@ bool InverseDynamics::terminate(const BlockInformation* blockInfo)
 bool InverseDynamics::output(const BlockInformation* blockInfo)
 {
     // Get the KinDynComputations object
-    auto kinDyn = getKinDynComputations(blockInfo).lock();
+    auto kinDyn = getKinDynComputations();
     if (!kinDyn) {
         wbtError << "Failed to retrieve the KinDynComputations object.";
         return false;
--- a/toolbox/CMakeLists.txt
+++ b/toolbox/CMakeLists.txt
@@ -76,7 +76,7 @@ configure_block(
             src/base/ConvertStdVector.cpp
             src/base/RobotInterface.cpp
             src/base/Signal.cpp
-            src/base/ToolboxSingleton.cpp
+            src/base/WholeBodySingleton.cpp
             src/base/WBBlock.cpp
     HEADERS include/base/Block.h
             include/base/BlockInformation.h
@@ -87,7 +87,7 @@ configure_block(
             include/base/ConvertStdVector.h
             include/base/RobotInterface.h
             include/base/Signal.h
-            include/base/ToolboxSingleton.h
+            include/base/WholeBodySingleton.h
             include/base/WBBlock.h
 )
 
--- a/toolbox/include/base/BlockInformation.h
+++ b/toolbox/include/base/BlockInformation.h
@@ -211,22 +211,6 @@ public:
      */
     virtual wbt::Signal getOutputPortSignal(const PortIndex idx,
                                             const VectorSize size = -1) const = 0;
-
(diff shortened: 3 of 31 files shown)


toolbox/src/InverseDynamics.cpp
```diff
@@ -139,20 +134,15 @@ bool InverseDynamics::initialize(BlockInformation* blockInfo)
     using namespace iDynTree;
 
     // Get the DoFs
-    const auto robotInterface = getRobotInterface(blockInfo).lock();
-    if (!robotInterface) {
-        wbtError << "RobotInterface has not been correctly initialized.";
-        return false;
-    }
-    const auto dofs = robotInterface->getConfiguration().getNumberOfDoFs();
+    const auto dofs = getRobotInterface()->getConfiguration().getNumberOfDoFs();
```
traversaro: Nitpick, not blocking the merging of the PR: I think that a good old `const size_t dofs` would be much more readable here and in other places.

francesco-romano:
Non going to stop this PR from merging, but just a couple of observations:
- Is [this](https://github.com/robotology/wb-toolbox/pull/115/commits/6908645763c8d820a68f2631f727212b43a5616f)  really needed? Why calling those function fail? I think this is something that should be changed in the implementation of the control board instead of adding workaround in dependant libraries
- the addition of `auto` to the language (C++) is a fantastic feature! Having said that, it should be prohibited almost everywhere. The code becomes almost unreadable. IMO, it should be used only to shorten the following kind of code:
```cpp
for (std::vector<MyType>::const_iterator it = ... )
// into
for (const auto& it = ... )
```
All the other uses should be banned.

diegoferigo:
> Is this really needed? Why calling those function fail? I think this is something that should be changed in the implementation of the control board instead of adding workaround in dependent libraries

I agree with you, this problem should be addressed upstream. However, generally speaking the toolbox should provide interfaces which should be ready to be used. Even if the issue wouldn't be there, it might be worth checking that the interface actually works before returning it. At least for debugging, if the interface never works at least you know it doesn't work from the first place if the error is spawned by `RobotInterface`. If, instead, the error is spawned by the block, this means that the interface has been correctly initialized and the error occurred later during its usage.

> The addition of auto to the language (C++) is a fantastic feature! Having said that, it should be prohibited almost everywhere.

IMO it is a trade off between readability and easiness of refactoring. The weight of one of these sides is just matter of the programmer. If done well, a small trade-off in readability provides a big potential during refactoring. I used `auto` here and there in addition to an extensive usage of curly braces initialization with this in mind. For simple types as this particular case, I like using auto because a switch from `unsigned` to `size_t` (@traversaro it is already in my list, along with a type for integers) would reflect in no warnings due to implicit conversions.

traversaro:
> Is this really needed? Why calling those functions fail? I think this is something that should be changed in the implementation of the control board instead of adding workarounds in dependant libraries

Calling those functions fails because the `remote_controlboard` device is waiting for the first read from the `stateExt:o` port. In the new Multiple Analog Sensors Interfaces I added a status to clearly reflect this ( https://github.com/robotology/yarp/blob/devel/src/libYARP_dev/include/yarp/dev/MultipleAnalogSensorsInterfaces.h#L32 ). 

As the details of the "being blocked until you received a valid read" can be quite tricky (which timeout do you use? how much time do you wait between a check and another? a read of a stateExt structure in which all the `_isValid` https://github.com/robotology/yarp/blob/master/src/libYARP_dev/src/devices/msgs/stateExt.thrift#L18 attribute are set to false counts as a valid read?) I think it is not crazy to leave this checks in the code that uses the `remote_controlboard` .

traversaro:
> For simple types as this particular case, I like using auto because a switch from unsigned to size_t (@traversaro it is already in my list, along with a type for integers) would reflect in no warnings due to implicit conversions.

Without sacrificing readability [`C4244`](https://msdn.microsoft.com/en-us/library/th7a07tz.aspx) and [`Wconversion`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) typically provide a warning in this case, even if a change of sign is not involved. I agree on the tradeoff between readability and easiness of refactoring, but please remember that code is read a lot of time by people that are not the maintainer.

Issue #119, opened 2018-05-23
Error compiling due to deprecation of control modes in YARP


hu-yue:
After updating YARP to the latest version, which deprecated the control modes ending with `2`, I got errors when compiling the toolbox, as below:

```
In file included from /usr/local/src/robot/yarp/src/libYARP_dev/include/yarp/dev/ControlBoardInterfaces.h:22:0,
                 from /usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/src/base/RobotInterface.cpp:10:
/usr/local/src/robot/yarp/src/libYARP_dev/include/yarp/dev/IControlMode2.h:19:81: error: conflicting declaration ‘typedef class yarp::dev::IControlMode yarp::dev::IControlMode2’
 YARP_DEPRECATED_TYPEDEF_MSG("Use yarp::dev::IControlMode instead") IControlMode IControlMode2;
                                                                                 ^
In file included from /usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/src/base/RobotInterface.cpp:1:0:
/usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/include/base/RobotInterface.h:20:15: note: previous declaration as ‘class yarp::dev::IControlMode2’
         class IControlMode2;
               ^
In file included from /usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/src/base/RobotInterface.cpp:11:0:
/usr/local/src/robot/yarp/src/libYARP_dev/include/yarp/dev/IControlLimits2.h:17:85: error: conflicting declaration ‘typedef class yarp::dev::IControlLimits yarp::dev::IControlLimits2’
 YARP_DEPRECATED_TYPEDEF_MSG("Use yarp::dev::IControlLimits instead") IControlLimits IControlLimits2;
                                                                                     ^
In file included from /usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/src/base/RobotInterface.cpp:1:0:
/usr/local/src/robot/robotology-superbuild/robotology/WBToolbox/toolbox/include/base/RobotInterface.h:23:15: note: previous declaration as ‘class yarp::dev::IControlLimits2’
         class IControlLimits2;
               ^
[ 23%] Building CXX object toolbox/CMakeFiles/WBToolbox.dir/src/YarpRead.cpp.o
[...]

PR #120, opened 2018-05-25
Switched to Yarp 3: updated interfaces and removed ConstString

diegoferigo:
Only minor changes are needed for the upcoming Yarp release. Fixes https://github.com/robotology/wb-toolbox/issues/119.

### DIFF
--- a/toolbox/CMakeLists.txt
+++ b/toolbox/CMakeLists.txt
@@ -5,7 +5,11 @@
 # FIND DEPENDENCIES
 # =================
 
-find_package(YARP REQUIRED)
+if(MSVC)
+    find_package(YARP REQUIRED)
+else()
+    find_package(YARP 2.3.73.3 REQUIRED)
+endif()
 
 # Fail if YARP is not compiled as shared library
 # see https://github.com/robotology/codyco-modules/issues/44
--- a/toolbox/include/base/RobotInterface.h
+++ b/toolbox/include/base/RobotInterface.h
@@ -21,11 +21,11 @@ namespace yarp {
         class IVelocityControl;
         class ITorqueControl;
         class IPWMControl;
-        class IControlMode2;
+        class IControlMode;
         class ICurrentControl;
         class IEncoders;
         class IMotorEncoders;
-        class IControlLimits2;
+        class IControlLimits;
         class IPidControl;
     } // namespace dev
 } // namespace yarp
@@ -156,7 +156,7 @@ public:
 // Specialize the getInterface template
 namespace wbt {
     template <>
-    bool RobotInterface::getInterface(yarp::dev::IControlMode2*& interface);
+    bool RobotInterface::getInterface(yarp::dev::IControlMode*& interface);
     template <>
     bool RobotInterface::getInterface(yarp::dev::IPositionControl*& interface);
     template <>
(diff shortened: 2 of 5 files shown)


traversaro:
I think turning the forward declaration to actual includes would permit to have a single branch of wb-toolbox compatible with both YARP 2.3.72 and YARP > 2.3.73, but that's your call. 

However, I think that enforcing the new minimum required version of YARP in https://github.com/robotology/wb-toolbox/blob/932c81cd049c7ce0c2494d60fe52bfd8c11dfd20/toolbox/CMakeLists.txt#L5 would definitely provide a clearer error in case of YARP/wb-toolbox version mismatch.

diegoferigo:
> However, I think that enforcing the new minimum required version of YARP would definitely provide a clearer error in case of YARP/wb-toolbox version mismatch.

I already thought of this, but Yarp devel still exports `2.3.73.3` and not 3.

traversaro:
> I already thought of this, but Yarp devel still exports 2.3.73.3 and not 3.

Exactly, while the master exports 2.3.72.* , so as YARP version configuration files use the `AnyNewerVersion` policy (https://github.com/robotology/yarp/blob/245690a2bb2b371a81efe0da2f9dbbd9f34e53b2/cmake/YarpDescribe.cmake#L82 , https://cmake.org/cmake/help/v3.8/module/CMakePackageConfigHelpers.html) having `find_package(YARP 2.3.73.3 REQUIRED)` will work fine.

diegoferigo:
I'd prefer to keep things clear and keep forward declarations. This PR breaks Windows running on Yarp `2.*` anyway since I removed `ConstString`s, and keeping the development of wb-toolbox aligned with Yarp as we do for many repos simplifies the maintenance (especially in the occasions of big changes like this coming release).

Said this, sadly appveyor will not pass until the next binary release :thinking: Considering we still do not have any test against Yarp devel on Windows, I'm going to merge anyway.

traversaro:
> Said this, sadly appveyor will not pass until the next binary release thinking Considering we still do not have any test against Yarp devel on Windows, I'm going to merge anyway.

Ok. Approving, but note that both this problems would not be present if we just switched to use includes in place of forward declaration.

traversaro:
We had a similar problem with forward declaration in gazebo-yarp-plugins as Gazebo was migrating from `boost::shared_ptr` to `std::shared_ptr`, see  https://github.com/robotology/gazebo-yarp-plugins/pull/207#discussion_r43463799 .

diegoferigo:
> note that this problem would not be present if we just switched to use includes in place of forward declaration

I see the point for the interfaces, not for the removal of `ConstString`. I didn't try, but I think that I should revert [this edit](https://github.com/robotology/wb-toolbox/pull/120/commits/807a696661ec1d9902218d6b0eed9e0a24a584d4#diff-563c9313c46a1d98fe8c205a9ec54756L271) to let windows build.

traversaro:
> I see the point for the interfaces, not for the removal of `ConstString`.

Yep, I mean for the interfaces. Removing the  `ConstString` is a totally unrelated change.

diegoferigo:
I gave it a try, but unfortunately despite `RobotInterface` might work, blocks that read those interfaces which are calling methods without the `2` do not work. Let's keep this broken status of Windows on current `yarp@v2.*` with `wb-toolbox@devel`. Note that this is an uncommon combination, hopefully it is just matter of a couple of weeks.

traversaro:
Yep, no problem!

diegoferigo:
Travis is clogged. Since Travis builds fine on my fork, I'm going to merge the PR.

PR #121, opened 2018-05-25
On Travis use devel branch for dependencies when devel is targeted

diegoferigo:


diegoferigo:
Travis is correctly failing during the test of the toolbox, but now the devel branch is cloned from the dependencies. Merging.

PR #124, opened 2018-05-28
Motor constants support (torque constant and back EMF)

### LINKED ISSUE #118: Provide a block for setting/getting motor control parameters(K_tau, K_bemf, K_ff)
**Desription**:
- Currently, we use `yarpmotorgui`, to set the motor control parameters, $K_\tau$,$K_{bemf}$, $K_{ff}$. It would be beneficial for many users if we can set/get the motor control parameters from the Simulink.

diegoferigo:
This PR adds to the _Set References_ and _Get Measurement_ block the capability of reading and setting motor's torque constant and back EMF. These values are usually use at firmware level.

Closes #118


toolbox/src/SetReferences.cpp
```diff
@@ -231,6 +264,26 @@ bool SetReferences::initialize(BlockInformation* blockInfo)
...
+        if (!getRobotInterface()->getInterface(iTorqueControl) || !iTorqueControl) {
+            wbtError << "Failed to get ITorqueControl interface.";
+            return false;
+        }
+        // Resize the vectors
+        pImpl->motorParams.resize(dofs);
+        pImpl->defaultMotorParams.resize(dofs);
+        // Get the default values
+        for (unsigned m = 0; m < dofs; ++m) {
+            if (!iTorqueControl->getMotorTorqueParams(m, &pImpl->defaultMotorParams[m])) {
```
traversaro: If I am not wrong,  this will cause a blocking RPC call (see https://github.com/robotology/yarp/blob/devel/src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp#L2613). I am not sure, but I suspect that the additional delay is not compatible with the real-time behavior of the controller.

toolbox/src/SetReferences.cpp
```diff
@@ -239,47 +292,57 @@ bool SetReferences::initialize(BlockInformation* blockInfo)
...
+    else if (pImpl->referenceType == impl::ReferenceType::MOTOR_BEMF
+             || pImpl->referenceType == impl::ReferenceType::MOTOR_KTAU) {
+        // Get the interface
+        ITorqueControl* interface = nullptr;
+        if (!robotInterface->getInterface(interface) || !interface) {
+            wbtError << "Failed to get ICurrentControl interface.";
+            return false;
+        }
+        for (unsigned m = 0; m < dofs; ++m) {
+            if (!interface->setMotorTorqueParams(m, pImpl->defaultMotorParams[m])) {
```
traversaro: See https://github.com/robotology/wb-toolbox/pull/124/commits/80273ce43ef0a0e7dab083772ced9f843bffc781#r191138043  but for the set (https://github.com/robotology/yarp/blob/devel/src/libYARP_dev/src/devices/RemoteControlBoard/RemoteControlBoard.cpp#L2597).

diegoferigo:
I see, thanks for bringing this up! What about documenting this behavior in the block documentation. Users usually do not change those values continuously, and the block can be disabled with an _Enabled Subsystem_ after have gathered / stored the first data.

traversaro:
> I see, thanks for bringing this up! What about documenting this behavior in the block documentation? Users usually do not change those values continuously, and the block can be disabled with an Enabled Subsystem after have gathered / stored the first data. 

Ok for me, if this is something reasonable for simulink users @DanielePucci @gabrielenava @vibhoraggarwal @S-Dafarra @nunoguedelha

traversaro:
Or perhaps Simulink users prefer a workflow more similar to SetLowLevelPIDs blocks?

S-Dafarra:
According to the modifications I saw for the implementation of gain scheduling, a similar feature can be implemented also for the torque loop. It means that in principle we could be able to change the gains several time when running the controller. From the interface point of view, the method to be called is always the same, while the user should specify also a smoothing time value to be set before hand.

All to say that it may be worthwhile to have a proper block which can call the setPID method at runtime if in a future we are planning to use the gain scheduling feature.

diegoferigo:
Closing in favor of https://github.com/robotology/wb-toolbox/pull/135.

PR #126, opened 2018-05-29
Signal const correctness

diegoferigo:
The last refactor of Signal which introduced templates for getting a setting the handled buffer introduced a subtle bug which allows to modify the buffer from const objects.

In fact, the function `T* getBuffer() const` was returning a pointer which could be used to change the buffer content.

This PR enforces the const correctness, and removes the former logic`Signal::isConst()`. The `BlockInformation` interface has been updated to return a `const Signal` for block's inputs.

### DIFF
--- a/toolbox/src/base/CoderBlockInformation.cpp
+++ b/toolbox/src/base/CoderBlockInformation.cpp
@@ -90,8 +90,8 @@ BlockInformation::VectorSize CoderBlockInformation::getOutputPortWidth(const Por
     return pImpl->outputPortDimensions.at(idx).at(1);
 }
 
-wbt::Signal CoderBlockInformation::getInputPortSignal(const PortIndex idx,
-                                                      const VectorSize size) const
+const wbt::Signal CoderBlockInformation::getInputPortSignal(const PortIndex idx,
+                                                            const VectorSize size) const
 {
     if (pImpl->inputSignals.find(idx) == pImpl->inputSignals.end()) {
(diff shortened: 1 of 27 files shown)


toolbox/src/base/CoderBlockInformation.cpp
```diff
@@ -90,8 +90,8 @@ BlockInformation::VectorSize CoderBlockInformation::getOutputPortWidth(const Por
     return pImpl->outputPortDimensions.at(idx).at(1);
 }
 
-wbt::Signal CoderBlockInformation::getInputPortSignal(const PortIndex idx,
-                                                      const VectorSize size) const
+const wbt::Signal CoderBlockInformation::getInputPortSignal(const PortIndex idx,
```
traversaro: I am not sure if changing the constness of a return value of a function changes anything at all, but if it working I do not think it is a big problem.

diegoferigo:
@traversaro I changed a few things. You were right about the returned value, users could have used a non-const Signal. Now I introduced new types for input and output signal as shared pointers, and in the former case BlockInformation returns a shared pointer to a const Signal.

This approach should solve the problem gracefully.

Disclaimer: I still have to test this solution is simulink. For the time being it just compiles fine.

traversaro:
Much better!

diegoferigo:
Everything looks working properly also in Simulink. This branch does not compile against yarp >= 3, but https://github.com/robotology/wb-toolbox/pull/133 which is based on top of this one fixes it.

Merging.

PR #130, opened 2018-06-04
Link the Yarp libraries dependencies, instead of using YARP_INCLUDE_DIRS

nunoguedelha:
Refer to the https://github.com/robotology/robotology-superbuild/issues/81 for more details.

### DIFF
--- a/toolbox/autogenerated/CMakeLists.txt
+++ b/toolbox/autogenerated/CMakeLists.txt
@@ -29,11 +29,13 @@ set_target_properties(ClockServer PROPERTIES POSITION_INDEPENDENT_CODE ON)
 # YARP
 find_package(YARP REQUIRED)
 
-# Link the dependencies libraries
-target_link_libraries(ClockServer YARP::YARP_OS)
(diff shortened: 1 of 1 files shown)


nunoguedelha:
@diegoferigo , actually that was my initial intention, but I thought the new way of using Yarp dependencies was to avoid extracting directly the interface includes path, except when really required, typically when we're using SWIG for generating bindings, like in the case of `yarp-matlab-bindings`.
Since we're not using SWIG here I was going directly for the libraries linking.

nunoguedelha:
Anyway, I'll following your suggestion.

nunoguedelha:
@diegoferigo , done.

diegoferigo:
> I thought the new way of using Yarp dependencies was to avoid extracting directly the interface includes path, except when really required

@nunoguedelha Yes you are right on this, but I think this is the cleanest way in this case since it is not strictly required linking any library. I prefer extracting a directory at CMake level rather than calling the linker without any reason. Thanks for the fix! :smile:

PR #133, opened 2018-06-13
Changes after Yarp 3 release

diegoferigo:
This PR introduces the following changes:

- Yarp 3 is now a dependency
- Included `clock_rpc.thrift` taken from gazebo-yarp-plugins
- Autogenerate the ClockServer class from the thrift file
- Use idyntree master in Travis


toolbox/thrift/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) 2018 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+# This software may be modified and distributed under the terms of the
+# GNU Lesser General Public License v2.1 or any later version.
+
+# GENERATE THE LIBRARY
+# ====================
+
+# TODO: sooner or later this file will be stored in YARP
+yarp_idl_to_dir(clock_rpc.thrift
```
traversaro: Is this working also with `ALLOW_IDL_GENERATION` set to `OFF` ?

traversaro: See https://github.com/robotology/yarp/issues/368 .

diegoferigo: @traversaro Not at the first run, and sadly that variable is disabled by default. Is [this policy](https://github.com/robotology/icub-main/issues/59#issuecomment-61812344) still valid? In this case, I should commit the autogenerated files.

traversaro: `yarp_idl_to_dir` is typically used with that policy (files generated in the source tree and committed), while `yarp_add_idl` ( https://github.com/robotology/yarp/blob/master/cmake/YarpIDL.cmake#L29 ) is used when the files are generated in the build directory.

PR #137, opened 2018-06-22
New CMake structure and plugin-based blocks architecture

diegoferigo:
## New CMake structure

After the previous big refactorings (https://github.com/robotology/wb-toolbox/pull/56, https://github.com/robotology/wb-toolbox/pull/77, https://github.com/robotology/wb-toolbox/pull/115) the wb-toolbox became more modular and these modules have been kept as independent as possible.

After these changes to the C++ side, this PR split the components at the CMake level in the following components:

| Component name | Description |
| --- | --- |
| **`Core`** | It contains core interfaces and classes such as `Block`, `Signal`, `BlockInformation`, `Log` |
| **`Mex`** | It contains the S-Function that generates the mex file and the implementation of `BlockInformation` for using block libraries with Simulink |
| **`Coder`** | Implementation of `BlockInformation` for using block libraries with Simulink Coder output |
| **`Library`** | It contains all the implementations of `Block` |

**`Core`**, **`Mex`** and **`Coder`** represent the real machinery of the toolbox (whose targets are named just _Toolbox*_ for the time being). They will represent the new project described in #83.

#### Simulink

Until now the mex generated from the S-Function was dynamically linked to the analogous of the **`Library`** component. The split introduced a problem due to the factory machinery. Since **`Mex`** is standalone, it doesn't know any information about the library containing the `wbt::Block` objects to allocate (they can also be different libraries) and hence the factory function was an unresolved symbol in the S-Function itself.

This PR introduces a new plugin-based approach to tackle this problem. The S-Function running in Simulink uses `dlopen` in a plugin fashion to get the `Block` object, which must know the name of the library file that contains it. This information is stored in the Simulink Library (`.slx` file).

#### Simulink Coder (C++)

Instead, the **`Library`** component can be linked either dynamically or statically with **`Coder`** and a main function in order to use it as a standalone C++ application after the code has been generated by Simulink Coder.

cc @traversaro @DanielePucci  @pattacini

PR #150, opened 2018-08-08
Improve CMake components

diegoferigo:
After #137 this PR polishes the usage of CMake components and cleans leftovers.

cc @drdanz


toolbox/src/sharedlibpp/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ set_target_properties(sharedlibpp PROPERTIES
 
 install(
     TARGETS sharedlibpp
-    EXPORT sharedlibpp
+    EXPORT sharedlibppExport
```
drdanz: I wouldn't touch `sharedlibpp`...

toolbox/src/sharedlibpp/CMakeLists.txt
```diff
@@ -43,5 +43,5 @@ include(InstallBasicPackageFiles)
 install_basic_package_files(sharedlibpp
     VERSION ${PROJECT_VERSION}
     COMPATIBILITY AnyNewerVersion
-    TARGETS sharedlibpp
+    EXPORT sharedlibppExport
```
drdanz: ...but if you change `sharedlibpp` you should pass `FIRST_TARGET` here.

CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ endif()
 # Add custom functions / macros
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
-find_package(YCM REQUIRED)
```
traversaro: Can you bump the robotology-superbuild YCM tag before the superbuild points to this commit? Otherwise this will fail.

cmake/AddGeneratedCodeLib.cmake
```diff
@@ -95,6 +72,10 @@ file(STRINGS ${CODER_DEFINES_FILE} CODER_DEFINES)
 # Model includes
 set(CODER_INCLUDES "${AUTOGEN_ABSPATH}")
 
+# TODO: figure out what is the best way to get Matlab_ROOT_DIR.
+#       Should we find it here, use an input variable, of inheriting it
+#       from the scope?
```
traversaro: I would say "inheriting it from the scope, but checking the `MATLAB_FOUND` variable or similar, and if it is not defined print a clear error sayng that one should call `find_package(Matlab ..)` before using this CMake macro.

toolbox/src/sharedlibpp/CMakeLists.txt
diegoferigo: Good catch, thanks!

CMakeLists.txt
diegoferigo: Yep thanks, Travis already reminded me :sweat_smile:

cmake/AddGeneratedCodeLib.cmake
diegoferigo: Added the check, thanks for the advice.

PR #152, opened 2018-08-10
Tutorial: how to create a simulink library from scratch

diegoferigo:
This new tutorial covers the steps of developing a C++ block with the core features of Whole Body Toolbox and explains how to create from scratch a Simulink Library for exposing it to Simulink.


diegoferigo:
cc @traversaro

doc/mkdocs/data/create_new_library.md
```diff
@@ -0,0 +1,644 @@
...
+
+#### S-Function
+
+There are many types of functions for implementing an algorithm wrapped by a Simulink Block. In Matlab terminology, these functions are referred as [S-Functions](https://it.mathworks.com/help/simulink/sfg/what-is-an-s-function.html). You can read more about the supported types in [What type of S-Function should you use](https://it.mathworks.com/help/simulink/sfg/what-type-of-s-function-should-you-use.html).
+
+In short S-Functions provide a sort of interface (through C callbacks) where a given number of functions need to be implemented. Simulink knows the name of these functions and calls them on demand.
+
+#### Block Mask
+
+A Simulink Block is just a square with input and output ports. It is possible to "mask" a Simulink Block in order to provide a user-friendly GUI that can be useful e.g. for setting Block parameters. The mask may contain buttons, dropdown menus, ...
```
traversaro: Missing full stop?

doc/mkdocs/data/create_new_library.md
```diff
@@ -0,0 +1,644 @@
...
+
+#### Shared library
+
+A shared library is an object that contains compiled code (functions, classes, ...) which cannot be executed standalone. They can be either _static_ or _dynamic_. Static shared libraries are meant to be embedded inside the executable that calls their symbols, instead the code of dynamic shared libraries is only referenced inside the executable and called when needed.
+
+For grasping better this difference, if multiple executables link against the same static library, the same code is embedded inside all of them, resulting in bigger executables and code duplication. A dynamic library object instead can be shared by multiple executables that need only to know their location in the filesystem and which symbols they provide.
```
traversaro: I always found "shared" used as a synonym of "dynamic", so I am a bit confused by the definition of  "static shared library". Furthermore also saying that "A library is an object" is a bit confusing, at least based on the typical definition of "object files" in linkers literature. Furthermore, it would be nice to point the interested reader to some resource on this, such as https://www.lurklurk.org/linkers/linkers.html .

doc/mkdocs/data/create_new_library.md
```diff
@@ -0,0 +1,644 @@
...
+
+#### Plugin Library
+
+A plugin library is a particular type of a dynamic shared library. An executable can load it either at load time, i.e. at the beginning, or at run time, i.e. when needed during the execution. The libraries that can be loaded during runtime are called plugins.
```
traversaro: at the beginning --> when the executable is launched

doc/mkdocs/data/create_new_library.md
```diff
@@ -0,0 +1,644 @@
...
+
+![EmptySFunctionBlock](images/EmptySFunctionBlock.png)
+
+What you need to know for the time being is that the name of the plugin library generated from the C++ code will be stored in the _S-function parameters_ field. We provide a generic **Level-2 MEX S-Function** that reads this parameter and uses the library name name to dynamically load the dynamic shared library that contains the block's code.
+
+After the library is found in the filesystem and successfully loaded, Simulink allocates an object of the C++ class associated to the block functionality. Again, this information (the class name) is passed in the _S-function parameters_ field.
+
+## Develop the C++ plugin
+
+You already learned that Blocks are no more than regular C++ classes. They are not an exception, in fact all the components discussed until now are mapped to C++ classes or interfaces (virtual classes).
```
traversaro: interface  (virtual classes) ---> interfaces (pure virtual classes)

traversaro: Or  perhabs https://en.cppreference.com/w/cpp/language/abstract_class is better name?

doc/mkdocs/data/create_new_library.md
diegoferigo: Yes, just _static library_ would suffice. Thanks for spotting it!

doc/mkdocs/data/create_new_library.md
diegoferigo: An etc was maybe missing :sweat_smile:

doc/mkdocs/data/create_new_library.md
diegoferigo: > interface (virtual classes) ---> interfaces (pure virtual classes)

I didn't get what you meant here, but I agree that using abstract class would be better.

diegoferigo:
@traversaro Updated the tutorial with your suggestions, thanks!

doc/mkdocs/data/create_new_library.md
traversaro: Neither the C++ standard nor the C++ book by Bjarne Stroustrup define the "virtual class" term. However, as far as I checked on the internet (see for example https://en.wikipedia.org/wiki/Virtual_class), "virtual classes" typical refer to any class some methods of which can be override.  Classes in which some method implementation is missing at all (as typically done in interfaces) are sometime called **pure** virtual classes.

PR #155, opened 2018-08-20
Cleanup

diegoferigo:
Cleaned directory structure


diegoferigo:
@traversaro Probably an error doing `git mv`, thanks! Fixed and merging.

Issue #153, opened 2018-08-14
Problem with Set Reference block in Velocity Control Mode


FabioBergonti:
During some experiments I've tried to control the robot (iCubGenova02) in speed but it didn't move.
So I've done a very simple test, I've applied a sine as input for only one joint but it didn't work.
Using `yarp motorgui` I'm able to move joints in speed, so it seems that the following simulink block doesn't work
![image](https://user-images.githubusercontent.com/38210073/44093054-9a1d3a5c-9fd2-11e8-8eb9-68fd2189c95f.png)

My WBtoolbox is currently update with [this commit](https://github.com/diegoferigo/WB-Toolbox/commit/44554cc743da733e74bebb24cc85f5bfea91efdb)

diegoferigo:
Does this happen also in Gazebo?
Do you have any kind of errors (either in Simulink or the terminal where Matlab was launched)?
Does the simulation stop or continue?

On that branch I tested only _Position_ and _Position Direct_.

FabioBergonti:
@diegoferigo @fiorisi

> Does this happen also in Gazebo?

No, in gazebo if I send a velocity signal the motors moves. But the simulator is not tuned, so the robot behaves in a strange way (also @GiulioRomualdi did some tests)
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/38210073/44107028-6c0d0682-9ff6-11e8-9607-df9921bfeebf.gif)

> Do you have any kind of errors (either in Simulink or the terminal where Matlab was launched)?

In matlab and simulink I didn't have errors. I didn't check the terminal.

> Does the simulation stop or continue?

It continued


> On that branch I tested only Position and Position Direct

A week ago I did also a test using *master* but it didn't work as well

diegoferigo:
Now `master` is quite old and works only with yarp 2. No one ever used velocity control as far as I know in the past and I am not surprised something was wrong.

If problems are only on the real robot with the new branch I fear we need to take some time for debugging them.

> For doing the tests on the real robot I've created a new branch where some .config have been modified in order to control iCub as R1

It would be nice having a Simulink model implementing a simple velocity control of a single joint. If changes of the configuration files are required, considering that the current status is not working, a PR targeting the gazebo model of robots-configuration is appreciated.

diegoferigo:
[...]

PR #156, opened 2018-08-20
SetReference block in Velocity control mode: set the reference acceleration

diegoferigo:
Similarly to the Position control model, the Velocity control mode should set the reference acceleration used internally by the trajectory generation of the interface implementation. This PR allows setting this parameter from the mask of the _SetReference_ block.

The default value is very high because usually the desired behavior is a perfect tracking of the velocity profile.

Maybe related to #153.


diegoferigo:
Merging this PR. If it doesn't fix #153 I'll handle it in a new one.

PR #159, opened 2018-08-23
Fix qpOASES block stopping behavior

### LINKED ISSUE #158: QP block stops the simulation if QP fails
- [x] I already updated to the latest version I can use
- [x] I already checked similar issues using the search box
- [x] I already checked the website for known troubleshooting
- [x] I already cleaned my environment (by removing the wb-toolbox install folder and installing it again)
- [x] I already excluded the possibility that system configuration files (e.g. yarp related) might be responsible
- [x] I am sure the environment variables are correct
- [x] I tried to reproduce the error in simulation (if it affects the real robot)

#### Description

`QP` block  is supposed **not** to stop the simulation if the QP fails and the option `stop the simulation when the solver fails` is unchecked. Instead, even if the option is unchecked the simulation stops when the `QP` block fails to find a solution:

#### Platform

Verified in simulation with `icubGazeboSim` and `iCubGazeboV2_5`.

#### Reproducibility

`git clone` the [whole-body-controllers](https://github.com/robotology/whole-body-controllers) repository, `git checkout` the branch `WBToolbox4`, verify all the dependencies are correctly installed and run the `momentumBasedYoga` simulation.

#### Screenshots

screenshot from 2018-08-22 17-20-17

screenshot from 2018-08-22 16-26-22

screenshot from 2018-08-22 16-50-08

#### Additional information

- OS: [Ubuntu 16.04]
- Version / Commit: [`v4.0`]
- Matlab Version: [ `2017b`]

diegoferigo:
The block was following the user choice about stopping only if the solver failed its initialization (either a cold initialization or hotstart).

However, if the primal solution could not be computed, the block was stopping in any case. This PR fixes this behavior.

cc @gabrielenava

Fixes #158

### DIFF
--- a/doc/release/v4.1.md
+++ b/doc/release/v4.1.md
@@ -1,0 +1,34 @@
+# WB-Toolbox (YYYY-MM-DD) Release Notes {[`#v4.1`](https://github.com/robotology/wb-toolbox/releases/tag/v4.1)}
+
+[Description]
+
+## Important Changes
+
+## New Features
+
+### `ToolboxCore`
+
+### `WBToolboxLibrary`
+
+### `WBToolboxMex`
+
+### `WBToolboxCoder`
+
+## Bug Fixes
+
+### `ToolboxCore`
+
+### `WBToolboxLibrary`
+
+- Fixed failing behavior of qpOASES block [#158](https://github.com/robotology/wb-toolbox/issues/158)
+
+### `WBToolboxMex`
+
+### `WBToolboxCoder`
+
+## Contributors
+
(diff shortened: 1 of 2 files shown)


traversaro:
@gabrielenava the meaning of the status output is the following: https://www.coin-or.org/qpOASES/doc/3.2/doxygen/Utils_8hpp.html#ad4d2df38ec673930c29b3417dcb05125 .

diegoferigo:
@gabrielenava Thanks for the test! Merging.

PR #168, opened 2019-01-30
Align WBT and BlockFactory

diegoferigo:
This PR contains the modification required to work with the current `master` branch of BlockFactory.


toolbox/library/CMakeLists.txt
```diff
@@ -2,182 +2,88 @@
...
+    list(APPEND WBT_BLOCKS "MinimumJerkTrajectoryGenerator")
+endif()
 
-if(WBT_USES_QPOASES)
-    configure_block(BLOCK_NAME "qpOASES"
-        GROUP "Utilities"
-        LIST_PREFIX WBT
-        SOURCES src/QpOases.cpp
-        HEADERS include/WBToolbox/Block/QpOases.h)
+if(${qpOASES_FOUND})
```
traversaro: With this change, qpOASES blocks are compiled even if `WBT_USES_QPOASES` is not enabled. I am afraid this may be confusing.

toolbox/library/CMakeLists.txt
```diff
@@ -2,182 +2,88 @@
...
+list(APPEND WBT_BLOCKS "DotJNu")
+list(APPEND WBT_BLOCKS "Jacobian")
+list(APPEND WBT_BLOCKS "ForwardKinematics")
+list(APPEND WBT_BLOCKS "RelativeTransform")
+list(APPEND WBT_BLOCKS "CentroidalMomentum")
+
+# Other
+list(APPEND WBT_BLOCKS "RealTimeSynchronizer")
+
+if(${ICUB_FOUND})
```
traversaro: See https://github.com/robotology/wb-toolbox/pull/168/files#r252674954 .

toolbox/library/CMakeLists.txt
traversaro: If you want this to be the behaviour (not a big fan of it to be honest), problably it would make sense to just remove those variables then.

toolbox/library/src/Jacobian.cpp
```diff
@@ -264,8 +263,8 @@ bool Jacobian::output(const BlockInformation* blockInfo)
     Map<MatrixXdiDynTree> jacobianRowMajor = toEigen(pImpl->jacobian);
     Map<MatrixXdSimulink> jacobianColMajor(
         output->getBuffer<double>(),
-        blockInfo->getOutputPortMatrixSize(OutputIndex::Jacobian).first,
-        blockInfo->getOutputPortMatrixSize(OutputIndex::Jacobian).second);
+        blockInfo->getOutputPortMatrixSize(OutputIndex::Jacobian).rows,
+        blockInfo->getOutputPortMatrixSize(OutputIndex::Jacobian).cols);
```
traversaro: 👍

toolbox/library/CMakeLists.txt
diegoferigo: Mmh I didn't realized I used the `_FOUND` here, my mistake. If you see in the bottom of this file, the `WBT_USES_` variables are used again. Thanks!

toolbox/library/src/Jacobian.cpp
diegoferigo: This was really necessary! I regretted using `pairs` and `tuples` in this case.

diegoferigo:
Done.

For what concern the temporary CI fix, I would keep it in this PR, regenerate the docker images after merging, and then revert the commit.

matlab/library/WBToolboxLibrary_repository.mdl
```diff
@@ -5837,14 +5837,14 @@ Library {
 	}
 	Block {
 	  BlockType		  S-Function
-	  Name			  "Yarp Read"
+	  Name			  "YarpRead"
```
traversaro: Even if WB-Toolbox  will always ship the manually written `.slx`/`.mdl`, I think this space removal is good for eventual consistency with blockfactory libraries generated by https://github.com/robotology/blockfactory/issues/10 .

diegoferigo: For sure less error prone, it was a trivial change, just some `sed` magic. Lucky us that we use mdl and not slx.

traversaro: OT: To be honest I am still confused on the role of `mdl` and `slx` in WBToolbox. The `.slx` model include `.mdl` one, but has support for the images? We could use the `mdl` also in the example of BlockFactory?

diegoferigo: I should try again substituting the `slx` with the `mdl` also in the installed tree.

diegoferigo: Ok I just tried. The models work fine, but the library does not show up in the Simulink Library.

screenshot

diegoferigo: I mean, if we really want to avoid having a binary file in our tree we can still install the mdl, and then the user will resave the slx the first time he / she opens the simulink library.

diegoferigo:
Merged, thanks.

matlab/library/WBToolboxLibrary_repository.mdl
diegoferigo: Some while ago I stored some information on the export process here (a bit unusual but there was no website at the moment):

https://github.com/robotology/wb-toolbox/blob/368119e32955e8eed5626ddc871bba76f4cd28c1/matlab/CMakeLists.txt#L9-L14

> why we do not run the export_library.m script as part of the build project to create the .slx in the build directory, and then we install as any build artefact?

Usually the `mdl` is exported with the matlab version of the maintainer of the library (:hand:). If users have an old version of Matlab they cannot convert a model / library created with a newer version.

traversaro: Cool. I finally got it, I think. So if we did not wanted to commit a `.slx` we could first save the mantainer's `.mdl` to a `mdl` compatible with 2014b: 
~~~
save_system(libraryName, 'WBToolboxLibrary', 'ExportToVersion', 'R2014B_MDL');
~~~
and then convert it to `R2014B_SLX` just during the build on the user's machine, before its install, to support correctly the `EnableLBRepository` option.

diegoferigo: Exactly

Issue #167, opened 2019-01-21
[Announcement] Script to port Simulink models from WBT v4 to WBT v5


diegoferigo:
The upcoming `v5` release of WBT (at the time of writing in the `devel` branch) will break the compatibility of the controllers based on the previous releases.

I developed a [bash script](https://gist.github.com/diegoferigo/e2a693b9329c26e7f90d17cf1c3f014e) to automatically update the controllers. It should work [OOTB](https://www.wikiwand.com/en/Out_of_the_box_(feature)) in both GNU/Linux and macOS, and on Windows using the shell provided by [Git BASH](https://gitforwindows.org/).

It should be properly tested by @gabrielenava, please let me know if you face any issue.

### Usage

```bash
wget https://gist.githubusercontent.com/diegoferigo/e2a693b9329c26e7f90d17cf1c3f014e/raw/974acd3c748884161fd70b39020fbf2b5aacae6b/WBT4_to_WBT5.sh
chmod +x WBT4_to_WBT5.sh
#
# Read the help executing:
# ./WBT4_to_WBT5.sh -h
#
# Example:
# ./WBT4_to_WBT5.sh -i model_WBT4.mdl -o model_WBT5.mdl
```

cc @traversaro

lrapetti:
While trying to port a model to `WBT` using this bash script I am getting the following error:
```
$ ./WBT4_to_WBT5.sh -i torqueBalancingPHRIStandup.mdl -o torqueBalancingPHRIStandup_WBT5.mdl
\e[1;32m==>\e[0m Processing torqueBalancingPHRIStandup.mdl model
  \e[1;34m->\e[0m Creating copy of the input mdl
  \e[1;34m->\e[0m Processing wholeBodyStates blocks
sed: 1: "torqueBalancingPHRIStan ...": undefined label 'orqueBalancingPHRIStandup_WBT5.mdl'
```

The model I am trying to port is the one in https://github.com/Yeshasvitvs/whole-body-controllers/tree/andy_standup.

diegoferigo:
Are you using the bash shell?

diegoferigo:
I suspect the error is related to https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux. Can you please check?

lrapetti:
> I suspect the error is related to https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux. Can you please check?

[...]

PR #169, opened 2019-02-01
Release v5

diegoferigo:
Ready to merge and release @traversaro


diegoferigo:
Since the `whole-body-controllers` still have to be ported to the new `WBT` version using #167, we should also checkout the pre-release commit in the superbuild. I will take care of it.

cc @gabrielenava

diegoferigo:
Hopefully Travis will be happy after commit https://github.com/robotology-playground/docker_images/commit/cb4ab232a327b03ae4c84fd9e3d2f0bb8a6fc213.

PR #172, opened 2019-02-04
Fix yarp::os::Network problems

### LINKED ISSUE #171: Behavior mismatch on yarp network between Simulink and autogenerated sources
I have a very simple Simulink model:

model

When it runs from Simulink, everything works fine. It is configured to move an iCub joint in Gazebo and it does it successfully. However, running the autogenerated sources fails the with the following error:

```
yarp: YARP not initialized; create a yarp::os::Network object before using ports
[Output]
/home/dferigo/git/WB-Toolbox/toolbox/library/src/SimulatorSynchronizer.cpp@output:187
Error connecting to simulator clock server.
```

I suspect something related to the calls of `yarp::os::Network`.

diegoferigo:
Instead of calling `init()` and `fini()` methods, it is better to store an instance of `yarp::os::Network` inside each class that needs yarp network access.

Fixes #171.

### DIFF
--- a/toolbox/base/src/RobotInterface.cpp
+++ b/toolbox/base/src/RobotInterface.cpp
@@ -54,6 +54,7 @@ struct YarpInterfaces
 class RobotInterface::impl
 {
 public:
+    std::unique_ptr<yarp::os::Network> network = nullptr;
     std::unique_ptr<yarp::dev::PolyDriver> robotDevice;
     std::shared_ptr<iDynTree::KinDynComputations> kinDynComp;
     YarpInterfaces yarpInterfaces;
@@ -144,8 +145,11 @@ public:
 
     bool initializeRemoteControlBoardRemapper()
     {
+        if (!network) {
+            network = std::make_unique<yarp::os::Network>();
+        }
+
         // Initialize the network
-        yarp::os::Network::init();
         if (!yarp::os::Network::initialized() || !yarp::os::Network::checkNetwork(5.0)) {
             bfError << "YARP server wasn't found active.";
             return false;
@@ -220,8 +224,6 @@ RobotInterface::~RobotInterface()
     if (pImpl->robotDevice) {
         pImpl->robotDevice->close();
     }
-    // Finalize the network
-    yarp::os::Network::fini();
 }
 
 // GET METHODS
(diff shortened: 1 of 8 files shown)


diegoferigo:
This targets `master` since it is an hotfix.

toolbox/base/src/RobotInterface.cpp
```diff
@@ -144,9 +145,12 @@ class RobotInterface::impl
...
     bool initializeRemoteControlBoardRemapper()
     {
+        if (!network) {
+            network = std::make_unique<yarp::os::Network>();
+        }
+
         // Initialize the network
-        yarp::os::Network::init();
-        if (!yarp::os::Network::initialized() || !yarp::os::Network::checkNetwork(5.0)) {
+        if (!network->initialized() || !network->checkNetwork(5.0)) {
```
francesco-romano: nitpick: these methods are static

diegoferigo: Yes, they are. However analogously to what happens with ROS, storing internally the network object automatically provides RAII. It is also recommended in the [YARP documentation](http://www.yarp.it/classyarp_1_1os_1_1Network.html#ac4eae967d4d518ec56632dfc0c9360f4).

francesco-romano: I am referring on how you changed the call to `checkNetwork`, `initialized` and `connect`. C++ allows to call a static method on an object, but I've always found it disturbing.

Regarding the RAII I agree, but I am sure there was a reason at the time I used the explicit `init` and `fini`. It might not being needed anymore now, though.

traversaro: I think that @francesco-romano suggests to call them as `yarp::os::Network::initialized()` , so it is more clear that they are actually static methods.

diegoferigo: Thanks, I got it now. I was thinking if it was necessary to check `pImpl->network != nullptr` and I ended up not doing it, but calling the methods from the singleton is cleaner and it does not have this problem.

francesco-romano: I don't think the constructor can fail if not for memory error. You can probably assert on it if you want. But I would not be too concerned for that.

diegoferigo: I thought the same :+1: Anyway, I substituted all the calls of the pimpl member with calls from the singleton. Now the pimpl member is only used to automatically call `init()` and `fini()`.

PR #173, opened 2019-02-09
Support to change the base link of the loaded model

### LINKED ISSUE #67: Define base link of the floating base robot
Right now the toolbox uses as base link the default base link set into `iDynTree::KinDynComputation`, which is the first link read from the urdf.

It could be worth adding in the configuration of the toolbox an additional entry that defines the base link, with the default value of root_link.

This change should be trivial:

* Add an additional option to `WBToolbox.Configuration.m`
* Add an additional member to `wbt::Configuration`
* Read in `wbt::WBBlock` the joint name as a string from the struct passed from Simulink
* Update the _Configuration_ block's mask

Blocked by https://github.com/robotology/idyntree/issues/404.

diegoferigo:
This PR adds the support to change the base link of the loaded urdf model. For instance, in the case of icub, the default base link is `root_link`. Now, the Configuration block allows passing the new base link:

screenshot_20190209_160311

I am still debating if it is better forcing the user to specify the base link by default, as in this case, or make it optional and give the possibility to change it using a check-button to enable and disable it (it would be more complex to implement in the mask). The second choice would provide an easier usage since most of the time there's no need to change the base link. But, as it was happening before, it kind of hides the choice of the default base link which is performed internally in iDynTree, and having it explicit increases the awareness of the user.

Closes #67 

cc @gabrielenava @traversaro


diegoferigo:
@gabrielenava Any news on this?

traversaro:
I  imagine this will break any model that uses the configuration block, right?

gabrielenava:
>@gabrielenava Any news on this?

I forgot to update this PR, I've actually already tested this feature with `iRonCub`. I didn't find issues. In particular, in the flying controller, the `base link` orientation is controlled. I verified that the controlled frame orientation became the one of the `chest` frame when switching the `base link` from the `root link` to the `chest`.  Therefore it seems to me that the base link was changed correctly. 

>I imagine this will break any model that uses the configuration block, right?

It will, but it requires only to add a field in the [WBToolbox configuration structure](https://github.com/robotology/whole-body-controllers/blob/master/controllers/floating-base-balancing-torque-control/app/robots/iCubGazeboV2_5/configRobot.m#L14).

traversaro:
>  But, as it was happening before, it kind of hides the choice of the default base link which is performed internally in iDynTree, and having it explicit increases the awareness of the user.

To be honest, if you load the model from URDF, the actual choice of the base link is uniquely determined by how the URDF model itself.

diegoferigo:
I'm reviving the PR after some while. It seems to be that the C++ was fully functional but the missing handling of the default value of the base link was a problem.

The best thing to do is maybe rolling back to the default behaviour when the *Base Link* entry of the mask is empty, and use its value when instead is defined.

Since this is likely a trivial change, is there any current user of the toolbox that wants to get acquainted with the code / development workflow? We can do it also together in pair programming if it can help. It will be quite didactic because it touches everything: C++, Matlab, and Mask. @robotology/iit-dynamic-interaction-control. If no one shows up, I will finalize it when I get some time.

DanielePucci:
I would say that among @gabrielenava @lrapetti @Yeshasvitvs and @CarlottaSartore there may be interested users. Please let me know who intends to carry on the task guys

CarlottaSartore:
> Since this is likely a trivial change, is there any current user of the toolbox that wants to get acquainted with the code / development workflow? 

@diegoferigo I would like to do it! I would need some help since I have never actively developed anything in wb-toolbox but, as you said, it is a good chance to learn :)

traversaro:
Is anyone interested in **using** this feature? If so, I would be happy to work on it and finally merge it, otherwise I do not see the point in putting effort in it.

fyi @gabrielenava @Giulero @HosameldinMohamed @FabioBergonti @lrapetti @VenusPasandi @CarlottaSartore @valentino-razza @fabiodinatale  @robotology/iit-artificial-mechanical-intelligence

gabrielenava:
no interest from me at the moment

PR #178, opened 2019-06-05
Add yarp rpc block to WBToolbox

Yeshasvitvs:
Currently, the WBToolbox has only yarp read and yarp write blocks which serve the purpose of reading and writing to yarp ports from simulink. It is very useful to have a block that will enable users to communicate with rpc ports from simulink directly.This PR adds a new `YarpRpc` block to WBToolbox

@diegoferigo


toolbox/library/include/WBToolbox/Block/YarpRpc.h
```diff
@@ -0,0 +1,57 @@
...
+namespace blockfactory {
+    namespace core {
+        class BlockInformation;
+    } // namespace core
+} // namespace blockfactory
+
+/**
+ * @brief The wbt::YarpRead class
+ *
+ * In addition to @ref block_parameters, wbt::YarpRead requires:
```
diegoferigo: `wbt::block::YarpRpc`

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright (C) 2018 Istituto Italiano di Tecnologia (IIT)
```
diegoferigo: 2019

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+bool YarpRpc::configureSizeAndPorts(BlockInformation* blockInfo)
+{
+    // PARAMETERS
+    // ==========
+
+    if (!YarpRpc::parseParameters(blockInfo)) {
+        bfError << "Failed to parse parameters.";
+        return false;
+    }
```
diegoferigo: This is required only if in this method you need some information from the parameters. Since this is not the case, you can remove this `if` statement.

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
...
+    return true;
+}
+
+bool YarpRpc::initialize(BlockInformation* blockInfo)
+{
+    if (!Block::initialize(blockInfo)) {
+        return false;
+    }
+
+    using namespace yarp::os;
```
diegoferigo: You can remove the `using` since you already use the full namespace below

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (!yarp::os::Network::initialized() || !yarp::os::Network::checkNetwork(5.0)) {
+        bfError << "YARP server wasn't found active.";
+        return false;
+    }
+
+    // Open a rpc client port
+    // TODO: Check if the client name can be improved
+    pImpl->clientPortName = "/rpcClient" + pImpl->serverPortName;
+
+    if (!pImpl->rpcClientPort.open(pImpl->clientPortName)) {
```
diegoferigo: I think you can open an anonymous port. I do not remember exactly how to do it, try with `...`

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    yarp::os::Bottle response;
+
+    double trigger = triggerSignal->get<double>(0);
+
+    if (static_cast<bool>(trigger) && !pImpl->rpcClientPort.isWriting()) {
+        pImpl->rpcClientPort.write(pImpl->rpcCommand, response);
+        bfWarning << response.toString().c_str();
+    }
```
diegoferigo: Is the `isWriting` an actual case that can happen? If all RPC calls are blocking, I think that every previous call to the `output` method of this block should have been already concluded.

If this can happen, probably I would split the two conditions, and if the port `isWriting` the block could return `false` and stop the simulation.

toolbox/library/include/WBToolbox/Block/YarpRpc.h
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ endif()
 # ===================
 
 # YARP blocks
+list(APPEND WBT_BLOCKS "YarpRpc")
```
traversaro: RPC is an acronym, perhaps it could make sense to always spell it as YarpRPC? On the other hand also YARP is an acronym, so it could make sense to leave as it is.

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Good call ;) it works with `...`

matlab/library/WBToolboxLibrary_repository.mdl
```diff
@@ -5934,12 +5932,94 @@ Library {
...
+	  SFcnIsStateOwnerBlock	  off
+	  Object {
+	    $PropName		    "MaskObject"
+	    $ObjectID		    296
+	    $ClassName		    "Simulink.Mask"
+	    Type		    "YARP Rpc"
+	    Description		    "This block behaves as the command 'yarp rpc'. \n\nIt takes two parameters:\n1) Rpc server port "
+	    "name to which user wants to send rpc commands. This is a string datatype.\n2) Rpc command which is the command us"
+	    "er wants to send. This is a string datatype.\n\nThis block has an input signal called Trigger. This is intended t"
+	    "o be a boolean datatype and when the signal is true the rpc command is sent to the rpc server port."
```
traversaro: I would explicitly specify that the block is **blocking the simulation** until a reply is received, as one could expect that a separate thread is spawned to handle the RPC communication.

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,219 @@
...
+    InputSignalPtr triggerSignal = blockInfo->getInputPortSignal(InputIndex::TriggerSignal);
+
+    if (!triggerSignal) {
+        bfError << "Input signal not valid.";
+        return false;
+    }
+
+    yarp::os::Bottle response;
+
+    double trigger = triggerSignal->get<double>(0);
```
traversaro: I am a bit confused. The Simulink docs mention that an input should be of boolean type, but here is the input is read as a double signal, and converted back to boolean. @diegoferigo you know what is going on? BF converts automatically signals of different types to doubles?

traversaro:
I am a bit confused by the "trigger" input to be honest. Naively, I would have expected that it activates the RPC call only after a rising edge, but perhaps I just more used to what they do in the PLC world, that is probably not the main target of the WB-Toolbox.

toolbox/library/src/YarpRpc.cpp
diegoferigo: Boolean signals have never been used nor tested. WBT already has the support but it should be tested more thoroughly.

I am not currently sure if in the end the low-level buffer is always double. I had a memory of finding a similar situation and I remember I left a comment, but I couldn't find where.

diegoferigo:
> it activates the RPC call only after a rising edge

The logic should be this one. @Yeshasvitvs probably you have to store a new variable in the pimpl containing the previous value of the trigger port.

traversaro:
What happens if the server is closed during the simulation? The simulation hangs forever or there is some kind of timeout? Related to this, would it make sense to have also an output that reports if the latest call was successful or not?

toolbox/library/CMakeLists.txt
Yeshasvitvs: As discussed with @diegoferigo we can leave it as `YarpRpc`

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: As discussed with @diegoferigo I will leave the `double` type for the time being and add a TODO to update it after testing boolean support more thoroughly.

diegoferigo:
> The simulation hangs forever or there is some kind of timeout?

Similarly to what we do in [YarpRead.cpp#L285-L293](https://github.com/robotology/wb-toolbox/blob/bbd3e52ce53a705498537bc99410c59619c46fb2/toolbox/library/src/YarpRead.cpp#L285-L293), a timeout would a nice feature. I am not sure though how to do it with RPC ports, I never did it personally.

> Related to this, would it make sense to have also an output that reports if the latest call was successful or not?

In my opinion if the RPC call fails, the `output()` method should return false and stop the simulation.

traversaro:
Ack, for me we can also add this robustness feature in later PR.

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

Yeshasvitvs:
@diegoferigo @traversaro 

I added the option of rising  edge trigger to be default and also added timeout check for the rpc write https://github.com/robotology/wb-toolbox/pull/178/commits/1b73adf611e1753ce2521afd8e2072c128aed133

toolbox/library/include/WBToolbox/Block/YarpRpc.h
```diff
@@ -0,0 +1,60 @@
...
+    ~YarpRpc() override;
+
+    unsigned numberOfParameters() override;
+    bool parseParameters(blockfactory::core::BlockInformation* blockInfo) override;
+    bool configureSizeAndPorts(blockfactory::core::BlockInformation* blockInfo) override;
+    bool initialize(blockfactory::core::BlockInformation* blockInfo) override;
+    bool terminate(const blockfactory::core::BlockInformation* blockInfo) override;
+    bool output(const blockfactory::core::BlockInformation* blockInfo) override;
+
+    bool sendRpcCommand();
```
diegoferigo: This should not be part of the public API. Maybe you can move it to the impl class?

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    pImpl->network = std::make_unique<yarp::os::Network>();
+
+    if (!yarp::os::Network::initialized() || !yarp::os::Network::checkNetwork(5.0)) {
+        bfError << "YARP server wasn't found active.";
+        return false;
+    }
+
+    // Open a rpc client port
+    pImpl->clientPortName = "...";
```
diegoferigo: This variable is not meaningful and it is also dangerous since it is not the real client port name. Can you move the string directly to the `open` method of the port?

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+        bfWarning << "Interrupting rpc port";
+        pImpl->rpcClientPort.interrupt();
+    }
+
+    yarp::os::Network::disconnect(pImpl->rpcClientPort.getName(), pImpl->serverPortName);
+
+    // Close the port
+    pImpl->rpcClientPort.close();
+
+    return true;
```
diegoferigo: What would happen if at this stage the remote is no longer available? If we handle the timeout in the `output` method, we should be sure that the `terminate` returns true too. What I mean, just double check that this is the case, I am not sure what the `isWriting` method is supposed to return in this case for instance.

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+    bool rpcStatus = true;
+
+    //TODO: Update to getting bool type after
+    //testing boolean support more thoroughly
+    double trigger = triggerSignal->get<double>(0);
+
+    if (pImpl->risingEdgeTrigger) {
+
+        if (trigger && !pImpl->previousTrigger) {
+
```
diegoferigo: No spaces between `if`s

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+bool YarpRpc::output(const BlockInformation* blockInfo)
+{
+    InputSignalPtr triggerSignal = blockInfo->getInputPortSignal(InputIndex::TriggerSignal);
+
+    if (!triggerSignal) {
+        bfError << "Input signal not valid.";
+        return false;
+    }
+
+    bool rpcStatus = true;
```
diegoferigo: This is really a detail. I would initialize this to `false`. Not a big change in the current implementation, but it might hide possible errors if the cases below in the future will be modified and they contain a bug.

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    // Initialize the time counter for the timeout
+    const double t0 = yarp::os::SystemClock::nowSystem();
+
+    while(!pImpl->rpcClientPort.write(pImpl->rpcCommand, response)) {
+
+        yarp::os::Time::delay(0.0005);
+        const double now = yarp::os::SystemClock::nowSystem();
+
+        if((now - t0) > pImpl->timeout) {
```
diegoferigo: In the case of `YarpRead`, the timeout is a mask parameter. Not a blocking comment, but if this value is hardcoded as it is I would create a constant variable in this cpp file to store the timeout value.

toolbox/library/src/YarpRpc.cpp
```diff
@@ -0,0 +1,257 @@
...
+            bfError << "Failed to send the rpc command to the server "
+                    << " with in the configured timeout";
+            return false;
+        }
+    }
+
+    bfWarning << response.toString().c_str();
```
diegoferigo: Leftover?

matlab/library/WBToolboxLibrary_repository.mdl
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Nope :grinning: left it consciously as this is the only output users see on the matlab terminal to see if things went fine with the rpc command e.g. `[ok]` is the rpc commands to WBD rpc server. I prefer to leave it for the time being and change it if we decide to handle the rpc response as an output signal from the block at a later stage.

traversaro: Mhh, that is  going to be messy if several RPC blocks are used. How are the users expected to know that the random `[ok]` appearing in the output is the RPC  response? If we need to have this, I would at least suggest to 1) Output as an info, not a warning 2) Have an option to disable it.

matlab/library/WBToolboxLibrary_repository.mdl
```diff
@@ -1,7 +1,7 @@
 Library {
   Name			  "WBToolboxLibrary_repository"
   Version		  9.1
-  SavedCharacterEncoding  "US-ASCII"
+  SavedCharacterEncoding  "UTF-8"
```
traversaro: Clearly this is the ouput of an automatically generated file, but just checking @diegoferigo, to you think it is ok?

traversaro:
> Typically this can be checked from the response of the RPC call, but as far as I know there is no standard reply we use (right @traversaro?).

Indeed, as far as I know there is  no standard "successful" reply.
However, if such "check for correct reply" is ever implemented, it probably would make sense to also support the case to report the failure just as an output instead of just closing the simulation: from my point of view, blockfactory/wb-toolbox are the equivalent of scripting bindings (such as the one for Python/MATLAB generated by SWIG) but for dataflow systems. How would you consider if a bindings function on an error stopped the scripting interpreter (or exited the main program with [`std:exit`](https://en.cppreference.com/w/cpp/utility/program/exit)) instead of reporting an error to its caller?

matlab/library/WBToolboxLibrary_repository.mdl
```diff
@@ -5934,12 +5932,111 @@ Library {
...
+	  SFcnIsStateOwnerBlock	  off
+	  Object {
+	    $PropName		    "MaskObject"
+	    $ObjectID		    296
+	    $ClassName		    "Simulink.Mask"
+	    Type		    "YARP Rpc"
+	    Description		    "This block behaves as the command 'yarp rpc'. Accordingly, this is a blocked call blocking the "
+	    "simulation until a reply is received.\n\nIt takes two parameters:\n1) Rising edge trigger which is enabled by def"
+	    "ault. Rpc command is sent only if a rising edge is detected. i.e a signal change from 0 to 1. If rising edge trig"
+	    "ger is unchecked, a trigger signal of 1 will send the rpc command.\n2) Rpc server port name to which user wants t"
```
traversaro: Again, not blocking, but a common way to refer to this two variants of "activation" used in the PLC  world are "edge-trigger" and "level-triggered". Perhaps it could make sense to use them also in WB-Toolbox docs, if appropriate. See Section 1 of https://www.plcopen.org/sites/default/files/downloads/creating_plcopen_compliant_function_block_libraries.pdf for an example.

traversaro:
>  How would you consider if a bindings function on an error stopped the scripting interpreter (or exited the main program with std:exit) instead of reporting an error to its caller?

To get an idea of how this things are typically handled in scenarios were dataflow programming languages are used, but "stopping everything" is not a viable solution as the context is not the one of simulation, we can take inspiration from the technologies used in industrial automation/manufacturing.  An example for IEC 61131-3's Function Block Diagram, that is the standard dataflow programming language for industrial [PLC](https://en.wikipedia.org/wiki/Programmable_logic_controller)), you can check the PLCOpen's Motion Control blocks, for example in https://www.plcopen.org/sites/default/files/downloads/application_examples_with_plcopen_motion_control.pdf . All the blocks have an `Error` and `ErrorID` outputs, that can be checked by the program that uses those blocks.

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: @traversaro I checked with @diegoferigo and there isn't any info logging in block factory. The warning and error logs are designed to be compatible with the simulink output. So, as at the moment there is not output signal based on the response, we can add optional additional "expected response" parameter to the YarpRpc mask and check the response against it. In case if they do not match, can notify the user directly in simulink using `bferror` logger. For the time being I will leave a `TODO` message and comment out the warning of the response.

matlab/library/WBToolboxLibrary_repository.mdl
Yeshasvitvs: > Again, not blocking,

@traversaro I am not following this clearly. If I understood well the rpc client call to rpc server is a blocking call right ? In that case the simulation will be blocked till you get a response, as mentioned in https://github.com/robotology/wb-toolbox/pull/178#discussion_r291093250

> common way to refer to this two variants of "activation" used in the PLC world are "edge-trigger" and "level-triggered".

I will update the block description with edge-triggered trigger as the default option and level-triggered as the other option if the edge-triggered is unchecked.

@diegoferigo

toolbox/library/src/YarpRpc.cpp
Yeshasvitvs: @diegoferigo if it is set to false and if the trigger is never high then the output method always returns false. This is why I set the default value to be true and to be modified to false in case of calling the `sendRpcCommand()` method.

toolbox/library/include/WBToolbox/Block/YarpRpc.h
Yeshasvitvs: Done

toolbox/library/src/YarpRpc.cpp
diegoferigo: You're right, I missed it.

matlab/library/WBToolboxLibrary_repository.mdl
diegoferigo: I was wondering the same when I saw the diff. I guess this is related to the differences in the locale of Yeshi's and my machine. I don't think it's going to be a problem, but cannot guarantee. Let's keep this in mind.

diegoferigo:
> However, if such "check for correct reply" is ever implemented, it probably would make sense to also support the case to report the failure just as an output instead of just closing the simulation

I discussed f2f the very same thing with @Yeshasvitvs. When the check of the reply will be implemented, its content will be part of the error message. This can be also useful to users and developers to read the RPC reply for debugging purpose, as an alternative to reporting it either in the stdout or simulink logger. And I mean, users can set a wrong reply just for the sake of generating the error to get the actual reply.

matlab/library/WBToolboxLibrary_repository.mdl
Yeshasvitvs: Resolving this conversation
@traversaro @diegoferigo

diegoferigo:
I'm resuming this stale PR. @Yeshasvitvs was there any specific reason why it got stale? Was it working fine or other minor modification were pending?

cc @lrapetti

Yeshasvitvs:
@diegoferigo The feature was functional and I used it in standup controller to give external wrench inputs in Gazebo through the plugin. We just need to decide and finalize on what final error messages we should display for users on the terminal.

traversaro:
Hi @Yeshasvitvs, given the that time has passed, the PR needs to be rebased on the top of the latest master (or devel). Note that given that there were modifications in the .mdl files this may not be as smooth as when we do the same process with C++ or other textual files, perhaps @nunoguedelha has some pointers w.r.t. to that.

matlab/library/WBToolboxLibrary_repository.mdl
traversaro: > > Again, not blocking,
> 
> @traversaro I am not following this clearly. If I understood well the rpc client call to rpc server is a blocking call right ? In that case the simulation will be blocked till you get a response, as mentioned in [#178 (comment)](https://github.com/robotology/wb-toolbox/pull/178#discussion_r291093250)

I just meant that my comment was not blocking, I was not referring to the call.

traversaro:
I guess that the comment https://github.com/robotology/wb-toolbox/pull/178#discussion_r291635265 is still open?

nunoguedelha:
> Hi @Yeshasvitvs, given the that time has passed, the PR needs to be rebased on the top of the latest master (or devel). Note that given that there were modifications in the .mdl files this may not be as smooth as when we do the same process with C++ or other textual files, perhaps @nunoguedelha has some pointers w.r.t. to that.

Simulink has a tool `visdiff` for comparing and merging models, both .mdl and .xls. That `visdiff` GUI has a button for merging the model files. I implemented in my environment a glue configuration for integrating the `visdiff` tool with git commands from a terminal, which makes the process much easier. Actually the integration of such MATLAB tools with external version control systems is extensively documented in MATLAB help center:
- https://www.mathworks.com/help/matlab/matlab_env/comparing-files-and-folders.html
- https://www.mathworks.com/help/matlab/ref/visdiff.html
- https://www.mathworks.com/help/matlab/matlab_prog/customize-external-source-control-to-use-matlab-for-comparison-and-merge.html
- @traversaro pointed out this one: https://www.mathworks.com/help/simulink/ug/resolve-conflicts.html

I had disseminated this around me, but not enough. I could do a presentation to the lab.

traversaro:
At this point, I guess we can close this PR, and anyone that is interested in getting it in WB-Toolbox can update the PR to not have conflicts. But there is no benefit in keeping it open if no one is working on it, what do you think?

fyi @robotology/iit-artificial-mechanical-intelligence 

fyi in particular @lrapetti, if this used anywhere in the Andy/Ergocub controllers we can try to find someone to work on it.

lrapetti:
> fyi in particular @lrapetti, if this used anywhere in the Andy/Ergocub controllers we can try to find someone to work on it.

Currently, it is not used anywhere. I think it was tested back in time in the controllers developed by @Yeshasvitvs in https://github.com/ami-iit/element_reactive-phri. I am ok with proceeding with closing the PR

traversaro:
Ok, thanks @lrapetti ! If anyone wants to put the effort in getting this PR in a state in which it can be merged, feel free to ping me for help.

Issue #189, opened 2020-06-13
Optimize BlockInitialization MATLAB function to reduce model compilation time 


traversaro:
#### Summary
It seems that the BlockInitialization MATLAB function takes a lot of  time to execute in some models, impacting a lot the time necessary to start a simulation. 

#### Motivation

Profiling in some Simulink models that use WB-Toolbox revealed that during Simulink model compilation ( https://de.mathworks.com/help/simulink/ug/simulating-dynamic-systems.html#f7-8245 ) .  In particular, it seems that the part that costs more in term of time is the following  loop: https://github.com/robotology/wb-toolbox/blob/114ac0e76068399c51150cb55c5922634d199ad3/matlab/%2BWBToolbox/BlockInitialization.m#L30-L44, that takes  ~100 ms for each iteration, and perform an interation for each nesting level of the block, arriving to a time ~1/2 seconds for really nested models, and a total compilation time of ~1 minute for a model with a lot of WB-Toolbox blocks. 

#### Additional context

Further profiling to understand  which line of https://github.com/robotology/wb-toolbox/blob/114ac0e76068399c51150cb55c5922634d199ad3/matlab/%2BWBToolbox/BlockInitialization.m#L30-L44 takes so much time may be needed.

traversaro:
> Further profiling to understand  which line of https://github.com/robotology/wb-toolbox/blob/114ac0e76068399c51150cb55c5922634d199ad3/matlab/%2BWBToolbox/BlockInitialization.m#L30-L44 takes so much time may be needed. 

I did not profile those blocks, so it is possible that I may be wrong, but **under the hypothesis that the bottleneck is the call to `regexp`**, perhaps  a simpler  strategy to identify the block strings that start with the value of `analyzedScope` and end with `/ImConfig` may  be used. For example, just checking if the string startsWith the value  `analyzedScope`  (via https://de.mathworks.com/help/matlab/ref/startswith.html) and ends with the `/ImConfig` (via https://de.mathworks.com/help/matlab/ref/endswith.html) may be more efficient.

traversaro:
cc @Giulero @diegoferigo @nunoguedelha @lrapetti @GiulioRomualdi  @S-Dafarra

diegoferigo:
Thanks for opening the issue and providing a recap of all what we discussed, this is very helpful!

[...]

PR #190, opened 2020-06-15
Improved compilation time of WBToolbox blocks by optimizing `Blockinitialization()`

nunoguedelha:
This is related to https://github.com/robotology/wb-toolbox/issues/189. Refer to those issues for full details.

Improved compilation time of WBToolbox blocks by optimizing `Blockinitialization()` code:
- In `Blockinitialization()` current implementation, we first retrieve the list of all the WBT
  blocks in the model. Then we look for the closest Config block, starting from the block's
  scope, and going up in the tree.
- The optimization consisted in retrieving only the Config blocks (name under mask is `ImConfig`)
  instead of the full list. the reduced list is enough for the second regexp, when looking for
  the closest Config, since only the lines containing `ImConfig` are supposed to match the regexp.
- This change is compatible with the support of multi-config models.

### DIFF
--- a/matlab/+WBToolbox/BlockInitialization.m
+++ b/matlab/+WBToolbox/BlockInitialization.m
@@ -16,8 +16,8 @@ try
     blockNameSplit = strsplit(currentBlock,'/');
     topLevel = blockNameSplit{1};
 
-    % Get all the blocks from the top level of the system
-    blocks_system = find_system(topLevel,'LookUnderMasks','on','FollowLinks','on');
+    % Get all the "Config" blocks from the top level of the system (name of block under mask matching 'ImConfig')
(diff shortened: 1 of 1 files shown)


traversaro:
Why "Draft"?

nunoguedelha:
I was still building superbuild with the changes. Although I know that the changed file is simply copied to the install folder, I wanted to have that check end to end, building, then running a compilation under Matlab.
But I'm having issues building the superbuild because of my recent update of homebrew I think.

Anyway we can move the PR to ready now, since it's a trivial change.

nunoguedelha:
Well, great team work with all of you! And thx @Giulero for the stats on the code lines.

@diegoferigo I'll push force a commit to remove the reference you mentioned.

diegoferigo:
Merged into `master` and aligned `devel`.

nunoguedelha:
CC @gabrielenava

PR #191, opened 2020-06-29
Remove deprecated and not used iDynTree headers

traversaro:
WBBlock.cpp included some headers of iDynTree that will be deprecated in iDynTree 2,  and in ~~general are not actually used~~ actually they are used, but for compatibility with future iDynTree it is better not too IWYU in this case, as it is possible to compile fine without including them even with iDynTree 1.

This should avoid problems like https://github.com/robotology/robotology-superbuild/issues/422 in the future.


diegoferigo:
Thanks @traversaro, if I recall these were iwyu suggestions. Removing them is harmless.

Issue #20, opened 2016-04-07
Centroidal Momentum Matrix


DanielePucci:
It would be very good if the toolbox could provide the users with the centroidal momentum matrix.

azadm:
I strongly agree on this

DanielePucci:
Hi @azadm, for the time being, when I have to evaluate the CMM, I basically do two things:

1. often I have to evaluate terms of the kind: CMM(q)v, and to do this, it suffices to call the function H(q,nu) = CMM(q)nu  with the right arguments;

2. You can build the matrix CMM by columns. To do this, call the function H(q,nu) = CMM(q)nu a number of times equal to the degrees of freedom you are working with. In particular, let c_i denote the ith column of the CMM. Then, one has c_i = CMM(q)e_i, where e_i = [0 0 . . 1 . . 0], where the "1" is at the ith position.

azadm:
Thanks @DanielePucci

diegoferigo:
The block `Centroidal Momentum` returns the matrix as a signal. See ef08455

traversaro:
The Centroidal Momentum Matrix is the "jacobian" of the centroidal momentum w.r.t. the robot velocity, and it is not implemented by commit https://github.com/robotology/WB-Toolbox/commit/ef084555f097eb63aaef87febbc6ebfb2ed5943e .

traversaro:
See http://www.ambarish.com/paper/20080707_Orin_Goswami_momentum_matrix_IROS2008.pdf and http://ai2-s2-pdfs.s3.amazonaws.com/c854/a692d56b5c1f3b3d0eb61ed4ac47b30c47cc.pdf for more details about the CMM.

francesco-romano:
Sorry, my fault

traversaro:
@DanielePucci just to let you know,  implementing this will probably require a bit of modifications in KinDynComputatuins and on the WB-Toolbox side, but nothing too difficult. It would be a perfect task for someone that needs to get in touch in a "vertical" way with the WB-Toolbox/BlockFactory/iDynTree "stack".

gabrielenava:
[...]

PR #194, opened 2020-07-17
Created CMM Simulink Block

CarlottaSartore:
As commented in issue #20 , This PR contains the implementation of the Simulink block calling the `iDynTree::KinDynComputations::getCentroidalTotalMomentumJacobian()` function implemented by @GiulioRomualdi .

C.C. @traversaro @gabrielenava


CarlottaSartore:
I saw that the CI failed, I have tried again the compilation from a brand new build and everything worked fine. 

I checked the output of the CI tests and I saw that the problem is in the iDynTree dependency i.e. 

```
../toolbox/library/src/CMM.cpp:176:18: error: 'using element_type = class iDynTree::KinDynComputations {aka class iDynTree::KinDynComputations}' has no member named 'getCentroidalTotalMomentumJacobian'; did you mean 'getCentroidalTotalMomentum'?
     ok = kinDyn->getCentroidalTotalMomentumJacobian(pImpl->CentroidalTotalMomentumMatrix);
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                  getCentroidalTotalMomentum
[26/30] Building CXX object toolbox/library/CMakeFiles/WBToolbox.dir/src/DiscreteFilter.cpp.o
```

Which branch is the CI using ? is it possible that the error is due to  "wrong" iDynTree version? 

In my configuration I have iDynTree branch devel commit 03a667417dcc4f5ac2ec881bad99cc24cea24947.

diegoferigo:
I think that the images in https://github.com/robotology-playground/docker_images should be updated, if we're lucky we just need to re-run its CI. Let's wait and then retrigger the tests of this PR.

toolbox/library/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ list(APPEND WBT_BLOCKS "Jacobian")
 list(APPEND WBT_BLOCKS "ForwardKinematics")
 list(APPEND WBT_BLOCKS "RelativeTransform")
 list(APPEND WBT_BLOCKS "CentroidalMomentum")
+list(APPEND WBT_BLOCKS "CMM")
```
diegoferigo, line 41: I would extend CMM to CentroidalTotalMomentumMatrix pretty much everywhere (class names, files names, etc)

toolbox/library/src/CMM.cpp
```diff
@@ -0,0 +1,202 @@
...
+// ===========
+
+CMM::CMM()
+    : pImpl{new impl()}
+{}
+
+CMM::~CMM() = default;
+
+
+
```
diegoferigo, line 65: I don't think that the clang-format of this repo allows multiple blank lines. Are you sure you're using it? If not, can you update the style of all the C++ file with that?

toolbox/library/CMakeLists.txt
CarlottaSartore: Perfect, I will apply this change right away!

toolbox/library/src/CMM.cpp
CarlottaSartore: Yes sure !

CarlottaSartore:
@diegoferigo  Thank you for your review! 

>  did you export the mdl and slx for the correct Matlab version we support?

For exporting the slx I have used the [export_library](https://github.com/robotology/wb-toolbox/blob/devel/matlab/export_library.m) script, instead the mdl version is 2019b, which Matlab version is the default one? 2014b ?

toolbox/library/CMakeLists.txt
CarlottaSartore: Done in commit 95fa7bcb51c3f942271b80777c4861f7eb814e3c

toolbox/library/src/CMM.cpp
CarlottaSartore: Done in commit 95fa7bcb51c3f942271b80777c4861f7eb814e3c

traversaro:
Just to understand, @CarlottaSartore you have addressed all the reviews of @diegoferigo, right?

CarlottaSartore:
> Just to understand, @CarlottaSartore you have addressed all the reviews of @diegoferigo, right?

Yes, the only thing missing, to the best of my knowledge,  regards the version of Matlab as per the following comment: 

> For exporting the slx I have used the export_library script (i.e. it is 2014b), instead, the mdl version is 2019b, as Matlab version is the default one? 2014b ?

The CI tests are not passing but it could due to https://github.com/robotology/wb-toolbox/pull/194#issuecomment-660163374

diegoferigo:
> For exporting the slx I have used the export_library script (i.e. it is 2014b), instead, the mdl version is 2019b, which Matlab version is the default one? 2014b ?

I think that executing the `export_library` CMake should suffice, it already takes care of using the right version of Matlab:

https://github.com/robotology/wb-toolbox/blob/3fd8eee7835e54a0497e6bd8a72a3307a244124d/matlab/CMakeLists.txt#L9-L14

For what concerns CI, I triggered the generation of the updated Docker images, let's see if they succeed. The pipeline is a bit rusty lately, I had not much time to dedicate and there was no one else that was maintaining it. Let's see if the building process succeeds, it's gonna take a while.

diegoferigo:
Can you please rebase this PR on top of current `devel`? CI should be fixed now.

CarlottaSartore:
@diegoferigo  I have updated all the file with the clang-format (using QT), I have rebase on top of devel and fix one leftover. of the reanming in the simulink library. 
The version of the `WBToolboxLibrary.slx `(inside the exported folder) is R2014b and has been created using the `export_library` script.

diegoferigo:
Perfect, thanks for your contribution @CarlottaSartore, it looks good! A last nit, can you please squash together the two commits with the slx file? Minimizing the changes to binary files is very important to avoid the size of the git project to diverge :)

PR #198, opened 2021-01-25
CI: Add Windows support and add periodic build

traversaro:
I added a Windows CI job. 

Furthermore, I noticed that the Linux and macOS jobs were failing, mostly due to environment changes (ycm and YARP do not have anymore a devel branch). For this reason, I switched all the jobs to use `master` branches, and added a period nightly job to avoid such regressions in the future.

### DIFF
--- a/.ci/install.sh
+++ b/.ci/install.sh
@@ -1,11 +1,14 @@
 #!/bin/sh
 set -e
 
+echo "Creating $HOME/git"
 mkdir $HOME/git
 export CXXFLAGS="-Wno-unused-command-line-argument"
 
 # Install YCM
+echo "Moving to $HOME/git"
 cd $HOME/git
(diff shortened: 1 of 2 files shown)


diegoferigo (approved):
Thanks @traversaro! Regarding the failure, I suspect they might be pretty recent since last October (3 months ago) CI was working fine, see https://github.com/robotology/wb-toolbox/pull/194. Thanks for updating!

traversaro:
> the failure, I suspect they might be pretty recent since last October (3 months ago) CI was working fine, see #194. Thanks for updating!

Yes, they are due to YARP deleting the devel branch, see https://github.com/robotology/community/discussions/372 .

traversaro:
Superseded by https://github.com/robotology/wb-toolbox/pull/212 .

PR #206, opened 2021-05-18
Change mantainer and specify it in the README

traversaro:
As discussed on Teams. 

FYI @DanielePucci


diegoferigo (approved):
Could you please also update [CODEOWNERS](https://github.com/robotology/wb-toolbox/blob/master/.github/CODEOWNERS)?

Issue #202, opened 2021-02-15
Set/expose qpOASES's setToReliable option?


traversaro:
#### Summary

The qpOASES QP solver with default options sometimes mark as unfeasible problems that are feasible, see https://github.com/coin-or/qpOASES/issues/85 . However, from https://github.com/coin-or/qpOASES/issues/85#issuecomment-778382108 and https://raw.githubusercontent.com/coin-or/qpOASES/master/doc/manual.pdf page 30, it seems that there is a `setToReliable` option that should make the QP solver more reliable and the expense of computational speed.

#### Motivation

If this option is present, could it make sense to set them in https://github.com/robotology/wb-toolbox/blob/master/toolbox/library/src/QpOases.cpp#L320 or expose them?

#### Additional context

Related issues:
* https://github.com/robotology/robotology-superbuild/issues/239

traversaro:
This issue may be interesting for anyone using the WB-Toolbox for solving QP problems. 
cc @nunoguedelha @diegoferigo @gabrielenava @Giulero @lrapetti @Yeshasvitvs

diegoferigo:
From my side this sounds good. I'd recommend current users that have the qpOASES block in their Simulink Models to try this change. The modification should be straightforward as @traversaro pointed out:

https://github.com/robotology/wb-toolbox/blob/99156c73c6b651502317722dc908ba7a9ba5f4ce/toolbox/library/src/QpOases.cpp#L320-L322

```cpp
qpOASES::Options problemOptions; 
problemOptions.setToReliable();  // < -- HERE
pImpl->sqProblem->setOptions(problemOptions); 
```

traversaro:
> 
> 
> This issue may be interesting for anyone using the WB-Toolbox for solving QP problems.
> cc @nunoguedelha @diegoferigo @gabrielenava @Giulero @lrapetti @Yeshasvitvs

I think I forgot @CarlottaSartore @VenusPasandi @valentino-razza in this list, sorry for that. 

[...]

PR #208, opened 2021-05-18
[WIP] QpOases: Switch to use reliable options

traversaro:
Fix https://github.com/robotology/wb-toolbox/issues/202 . As per @lrapetti tests, on typical QP instances used in robotic Whole-Body Control the time difference is not big, and using the reliably settings seems safer.

### DIFF
--- a/toolbox/library/src/QpOases.cpp
+++ b/toolbox/library/src/QpOases.cpp
@@ -318,7 +318,7 @@ bool QpOases::initialize(BlockInformation* blockInfo)
 
     // Setup safe options
     qpOASES::Options problemOptions;
-    problemOptions.setToDefault();
+    problemOptions.setToReliable();
     pImpl->sqProblem->setOptions(problemOptions);
 
 #ifdef NDEBUG


traversaro:
cc @CarlottaSartore It could be worth to give it a shot to this PR.

diegoferigo (approved):

lrapetti:
As I was mentioning in https://github.com/robotology/wb-toolbox/issues/202#issuecomment-785259665, I have observed some unexpected behaviour. I would try to run again this option before merging (I will switch to this branch today).

traversaro:
> 
> 
> As I was mentioning in [#202 (comment)](https://github.com/robotology/wb-toolbox/issues/202#issuecomment-785259665), I have observed some unexpected behaviour. I would try to run again this option before merging (I will switch to this branch today).

Ah cool, I have missed that.

CarlottaSartore:
In my environment the change to reliable causes the QP to fail I would say always

traversaro:
As this seems to have a serious impact in some applications, and in any case a replacement block is being provided in https://github.com/robotology/wb-toolbox/pull/209 , I would close this.

PR #209, opened 2021-05-18
Add OSQP block

traversaro:
This PR add an OSQP block based on osqp-eigen to fix https://github.com/robotology/wb-toolbox/issues/180 . It requires `OsqpEigen` to be installed and the `WBT_USES_OSQPEIGEN` option to be set.

Known problems: 
* [x] KP1: The library is not visualized anymore in the Simulink Library Browser 

Solved problems:
* [x] KP2:  The `UseLb` and `UseUb` options are not supported
* [x] KP3: Setting only one of `UseLba`  and `UseUba` to true while the other to false, gives the wrong results
* [x] KP4: All the use of the bfError macro are commented out because their use result in a compilation error
* [x] KP5:  The `ComputeObjVal, `StopWhenFails` options are not supported
* [x] KP6: Non-deterministic "[OsqpEigen::Solver::initSolver] Unable to setup the workspace."  error during initialization

Basic Tests again qpOASES block:
sqp_comparison

Example model: 
[qpoases_vs_osqp_test_v3.zip](https://github.com/robotology/wb-toolbox/files/6528910/qpoases_vs_osqp_test_v3.zip)

PR #211, opened 2021-05-18
Re-implement the Holder block to make it compatible with code generation and 'treat as atomic unit'

### LINKED ISSUE #205: Re-implement the Holder block to make it compatible with code generation and "treat as atomic unit"
#### Summary

image

Re-implement the Holder block (in "Utilities") in order to avoid the use of:
- persistent variables, which are incompatible with code generation.
- continuous blocks (the clock Simulink block): a sub-system combining discrete and continuous blocks cannot be treated as an atomic unit.

#### Motivation

Treating sub-systems as atomic units speed up the simulation and allow the use of different sampling times on multiple sub-systems without rate conversion problems.
Code generation increases the compilation time but improves significantly the runtime performance.

#### Additional context

The [whole-body-controllers/controllers/floating-base-balancing-torque-control](https://github.com/robotology/whole-body-controllers/tree/devel/controllers/floating-base-balancing-torque-control) model uses at least 7 Holder blocks.

nunoguedelha:
Fixes #205 .


traversaro:
@nunoguedelha do you think we can wait a few days for this? For some internal projects, https://github.com/robotology/wb-toolbox/pull/209 may need to have an higher priority and I guess there are incompatibilities there.

traversaro:
> 
> 
> @nunoguedelha do you think we can wait a few days for this? For some internal projects, #209 may need to have an higher priority and I guess there are incompatibilities between this two PRs.

Actually I think we can wait a few more days for https://github.com/robotology/wb-toolbox/pull/209, and in practice it is blocked by https://github.com/robotology/robotology-superbuild/issues/719 . So I think we are good to do on this, not sure if @Giulero is able to quickly check and approve the PR or he is too busy with the experiments.

nunoguedelha:
> I guess there are incompatibilities between this two PRs.

Hi @traversaro , when you say "incompatibilities" you just mean that both PRs are changing the same `.slx` file, and **not** real incompatibilities in the implementation, right?

traversaro:
> 
> 
> > I guess there are incompatibilities between this two PRs.
> 
> Hi @traversaro , when you say "incompatibilities" you just mean that both PRs are changing the same `.slx` file, and **not** real incompatibilities in the implementation, right?

Exactly.

nunoguedelha:
> Exactly.

For the record, in this case the merge should be easy.

Giulero:
Today I could check the PR, testing some of our simulink models. They have a lot of holder blocks!

Giulero:
I run a model with the simulink simulator. The controller block (containing some holders) is set as an atomic unit running 10 times slower than the simulink solver.
The only error was related to the use of the clock (simulink) block (that I guess was the source of the issue also for the wbt holder block). Once that block was removed everything run smoothly!

Thanks Nuno!

Giulero (approved):

traversaro (approved):

traversaro:
@nunoguedelha I think we can merge this, what do you think? Then I would probably need to rebase https://github.com/robotology/wb-toolbox/pull/209 on top of the new master after this PR is merged, not sure if you have any advice on how to do this.

nunoguedelha:
> @nunoguedelha I think we can merge this, what do you think? Then I would probably need to rebase #209 on top of the new master after this PR is merged, not sure if you have any advice on how to do this.

Ciao @traversaro , sorry just saw this, had an afternoon a bit problematic. Yes we can merge this now. For rebasing your branch `fix/180` impacting the same library, it shouldn't be a problem. Here's what I would do:
1. rebase your branch on top of the new master, resolving the conflict on the lib files `WBToolboxLibrary_repository.mdl` and `WBToolboxLibrary.slx` choosing your version only, discarding mine.
2. create a new commit after coverting the libraries slx and mdl to version 2014b.
3. merge the libraries (manually or through the Simulink tool `visdiff`), then `git commit --amend`.

**I guess you main concern was more precisely how to merge the mdl and slx files.** It should be trivial since the blocks impacted by the 2 branches are distinct. I only changed the "holder" block in the "utilities". 

- Manually:
    1. create a copy of my library version to an untracked file that you can still see when you go back to to your branch.
    2. replace your "holder" block in the library by mine.
- With Simulink tool `visdiff`:
    1. After the initial merge (in favour of your version), and while still on your branch checkout, run a diff of the two versions of each file.
    2. A diff GUI opens in anew MATLAB process. There is a 'merge' button in the diff GUI you can use to merge the files.
- Using  "il Nuno" for doing the merge (this is a joke for @giulero): I already have the tool `visdiff` integrated with Git, so it's really easy for me. Ideally I should integrate it in WBToolbox and present it to the team, it's really useful.

nunoguedelha:
> The only error was related to the use of the clock

Yes @Giulero , indeed the clock used before this PR was the only thing preventing the "set as atomic unit", not the `persistent` instruction. The problem with the `persistent` instruction is the following:
- You use it in a MATLAB System function.
- You then can only use the "MATLAB Inerpreted" compilation and not "Code Generation".

But in general I think it's a good practice to avoid `persistent` as it tends not to be compatible with other features in Simulink, and most of all is error prone as it is a static variable that needs to be cleared via `clear functions` (if I'm not wrong), and for which `clear all` is uneffective.

nunoguedelha:
I'm not authorized to merge this PR. @traversaro I let you do it. I suggest a "Rebase and merge" since there is a single commit in my branch.

traversaro:
> I'm not authorized to merge this PR. @traversaro I let you do it. I suggest a "Rebase and merge" since there is a single commit in my branch.

Probably it was an artifact of broken CI.

traversaro:
> 
> 
> > @nunoguedelha I think we can merge this, what do you think? Then I would probably need to rebase #209 on top of the new master after this PR is merged, not sure if you have any advice on how to do this.
> 
> Ciao @traversaro , sorry just saw this, had an afternoon a bit problematic. Yes we can merge this now. For rebasing your branch `fix/180` impacting the same library, it shouldn't be a problem. Here's what I would do:
> 
>     1. rebase your branch on top of the new master, resolving the conflict on the lib files `WBToolboxLibrary_repository.mdl` and `WBToolboxLibrary.slx` choosing your version only, discarding mine.
> 
>     2. create a new commit after coverting the libraries slx and mdl to version 2014b.
> 
>     3. merge the libraries (manually or through the Simulink tool `visdiff`), then `git commit --amend`.
> 
> 
> **I guess you main concern was more precisely how to merge the mdl and slx files.** It should be trivial since the blocks impacted by the 2 branches are distinct. I only changed the "holder" block in the "utilities".
> 
>     * Manually:
>       
>       1. create a copy of my library version to an untracked file that you can still see when you go back to to your branch.
>       2. replace your "holder" block in the library by mine.
> 
>     * With Simulink tool `visdiff`:
>       
>       1. After the initial merge (in favour of your version), and while still on your branch checkout, run a diff of the two versions of each file.
>       2. A diff GUI opens in anew MATLAB process. There is a 'merge' button in the diff GUI you can use to merge the files.
> 
>     * Using  "il Nuno" for doing the merge (this is a joke for @Giulero): I already have the tool `visdiff` integrated with Git, so it's really easy for me. Ideally I should integrate it in WBToolbox and present it to the team, it's really useful.

Thanks @nunoguedelha ! This is quite interesting (and quite documented it seems: https://www.mathworks.com/help/simulink/ug/resolve-conflicts.html), I will try myself to familiarize with the tooling. fyi @pattacini

pattacini:
>Thanks @nunoguedelha ! This is quite interesting (and quite documented it seems: https://www.mathworks.com/help/simulink/ug/resolve-conflicts.html), I will try myself to familiarize with the tooling. fyi @pattacini

As @nunoguedelha was mentioning, this tooling helps quite a lot when the changes impact different parts of the model. When changes impact the same block instead, for example, vsdiff starts quickly becoming difficult to employ.

Issue #207, opened 2021-05-18
Website visualization is broken


traversaro:
Website is broken: https://robotology.github.io/wb-toolbox/mkdocs/index.html .

![website](https://user-images.githubusercontent.com/1857049/118620551-e39ba200-b7c5-11eb-8d6e-955a2133a748.PNG)

traversaro:
cc @diegoferigo

traversaro:
Now the job for generating the website is broken:
~~~
2021-07-08T06:47:33.1767736Z Patching output file 237/240
2021-07-08T06:47:33.1769143Z Patching output file 238/240
2021-07-08T06:47:33.1770006Z Patching output file 239/240
2021-07-08T06:47:33.1770432Z Patching output file 240/240
2021-07-08T06:47:33.1770972Z lookup cache used 983/65536 hits=7080 misses=1020
2021-07-08T06:47:33.1771416Z finished...
2021-07-08T06:47:33.1812383Z [100%] Built target dox
2021-07-08T06:47:33.2006450Z [100%] Generating doxygen documentation in /home/runner/work/wb-toolbox/wb-toolbox/build/doc/doxygen
2021-07-08T06:47:33.4654505Z [34mDEBUG    -  [0mLoading configuration file: /home/runner/work/wb-toolbox/wb-toolbox/doc/mkdocs/mkdocs.yml
2021-07-08T06:47:33.5660329Z [34mDEBUG    -  [0mLoaded theme configuration for 'material' from '/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/material/mkdocs_theme.yml': {'language': 'en', 'direction': None, 'features': [], 'palette': {'primary': None, 'accent': None}, 'font': {'text': 'Roboto', 'code': 'Roboto Mono'}, 'icon': None, 'favicon': 'assets/images/favicon.png', 'include_search_page': False, 'search_index_only': True, 'static_templates': ['404.html']}
2021-07-08T06:47:33.5665581Z [33mWARNING  -  [0mConfig value: 'site_url'. Warning: This option is now required. Set to a valid URL or an empty string to avoid an error in a future release.
2021-07-08T06:47:33.5667900Z [33mWARNING  -  [0mConfig value: 'site_url'. Warning: The 'use_directory_urls' option has been disabled because 'site_url' contains an empty value. Either define a valid URL for 'site_url' or set 'use_directory_urls' to False.
[...]

PR #212, opened 2021-05-19
Add conda-based Continuous Integration job

traversaro:
This PR adds a conda-based CI workflow for Windows, macOS and Linux, simpler then the already present (and currently broken) CI jobs.


traversaro:
The conda-based CI script seems to be working fine. As it seems easier to maintain that the apt/brew/vcpkg one, I would simply delete the `ci.yml` job and just use the `conda-ci.yml`, what do you think @diegoferigo ?

.github/workflows/conda-ci.yml
```diff
@@ -0,0 +1,72 @@
...
+        # Actual dependencies
+        mamba install eigen blockfactory idyntree osqp-eigen qpoases yarp
+        
+    - name: Dependencies
+      shell: bash -l {0}
+      run: |
+        # Compilation related dependencies 
+        mamba install cmake compilers ninja pkg-config
+        # Actual dependencies
+        mamba install eigen blockfactory idyntree osqp-eigen qpoases yarp
```
diegoferigo, line 44: Copy-and-paste duplication?

.github/workflows/conda-ci.yml
```diff
@@ -0,0 +1,72 @@
...
+      shell: bash -l {0}
+      run: |
+        cd build
+        cmake --build . --config ${{ matrix.build_type }}
+        
+    - name: Test
+      shell: bash -l {0}
+      run: |
+        cd build
+        ctest --output-on-failure -C ${{ matrix.build_type }}
```
diegoferigo, line 72: I don't think we have tests here, maybe we can leave these lines commented out?

diegoferigo (approved):
> The conda-based CI script seems to be working fine. As it seems easier to maintain that the apt/brew/vcpkg one, I would simply delete the `ci.yml` job and just use the `conda-ci.yml`, what do you think @diegoferigo ? Compilation via apt will be checked in any case daily via the https://github.com/robotology/robotology-superbuild .

I completely agree on using conda! We can then create a new workflow for just the website (or extend this one) as soon as https://github.com/robotology/wb-toolbox/issues/207 is fixed.

.github/workflows/conda-ci.yml
traversaro: Fixed.

.github/workflows/conda-ci.yml
traversaro: This line works fine also for repo that do not have tests, and will automatically start to run tests if tests are added. Given that I plan to wrap this logic in a github action to avoid duplicate code, and I would keep this for consistency with the rest of other repos.

traversaro:
> 
> 
> > The conda-based CI script seems to be working fine. As it seems easier to maintain that the apt/brew/vcpkg one, I would simply delete the `ci.yml` job and just use the `conda-ci.yml`, what do you think @diegoferigo ? Compilation via apt will be checked in any case daily via the https://github.com/robotology/robotology-superbuild .
> 
> I completely agree on using conda! We can then create a new workflow for just the website (or extend this one) as soon as #207 is fixed.

I think the workflow for the website is already there: https://github.com/robotology/wb-toolbox/blob/add-conda-ci/.github/workflows/website.yml ?

PR #227, opened 2021-11-26
Fix website generation by moving index.html of the documentation website in main repo

traversaro:
Fix https://github.com/robotology/wb-toolbox/issues/207 . 

In particular, this fixes a regression for which index.html files were never updated, probably with the intention of making sure that the index.html contained in the `gh-pages`  (added in https://github.com/robotology/wb-toolbox/commit/8ed7bc06a50bb9c07a7df2a88e69c422d1907c9d#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051) was never overwritten. To avoid this, the `index.html` is now contained in the `doc` directory, and copied in the deploy folder when necessary. This also ensure that everything necessary to generate the documentation website is managed and versioned as part of the main repo.

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -74,6 +74,7 @@ jobs:
 
       - name: Populate deploy folder
         run: |
+          cp doc/index.html $DEPLOY_FOLDER/index.html
           rm -rf $DEPLOY_FOLDER/mkdocs
           rm -rf $DEPLOY_FOLDER/doxygen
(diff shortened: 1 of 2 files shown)


diegoferigo (approved):
Thanks @traversaro for the fix, #207 remained in my backlog for long time. Checking the change, this is actually the approach used in other projects where the index is always copied instead of just being pushed the first time in the `gh-pages` branch. I guess this was a old inheritance, this website was AFAIK one of the first ones we had that was not using the ugly doxygen style :)

Issue #231, opened 2022-02-03
Solve osqp-eigen warning


traversaro:
Solve osqp warning:
~~~
/__w/robotology-superbuild/robotology-superbuild/src/WBToolbox/toolbox/library/src/OSQP.cpp: In member function 'virtual bool wbt::block::OSQP::output(const blockfactory::core::BlockInformation*)':
/__w/robotology-superbuild/robotology-superbuild/src/WBToolbox/toolbox/library/src/OSQP.cpp:549:50: warning: 'bool OsqpEigen::Solver::solve()' is deprecated: Use solveProblem() instead. [-Wdeprecated-declarations]
     bool solveReturnVal = pImpl->sqSolver->solve();
                                                  ^
~~~

I think it was introduced by osqp-eigen 0.7.0 .

YYB666:
I also meet the same warning, have you solved it?

traversaro:
No. However, it is just a warning, it should not prevent you from using the library.

Issue #216, opened 2021-05-24
Re-run clang-format on source files


traversaro:
See https://github.com/robotology/wb-toolbox/pull/209#discussion_r637747572 .

PR #234, opened 2022-10-18
add adaptive_rho, polish and maximum iteration options for OSQP

VenusPasandi:
With this PR, we add some options to the osqp block, namely:

- the flag for enabling and disabling the adaptive rho algorithm,
- the flag for enabling and disabling the polish algorithm,
- the maximum number of iterations


toolbox/library/src/OSQP.cpp
```diff
@@ -298,7 +305,7 @@ bool wbt::block::OSQP::solverInitialization(const BlockInformation*blockInfo)
     if (pImpl->useLb || pImpl->useUb) {
         pImpl->numberOfTotalConstraints += pImpl->numberOfVariables;
     }
-
+
```
traversaro, line 308: Nitpick: do not add trailing whitespace (see https://stackoverflow.com/questions/21410075/what-is-trailing-whitespace-and-how-can-i-handle-this).

toolbox/library/src/OSQP.cpp
```diff
@@ -318,14 +325,21 @@ bool wbt::block::OSQP::solverInitialization(const BlockInformation*blockInfo)
     // Setup options
     pImpl->sqSolver->settings()->setVerbosity(false);
 
-    // Set adaptive_rho to false to avoid converge problems
+    // Setup adaptive_rho
+    // Note that enabling adaptive_rho could cause converge problems
     // See https://github.com/oxfordcontrol/osqp/issues/151
-    pImpl->sqSolver->settings()->setAdaptiveRho(false);
```
traversaro, line 323: Interesting! I did not notice that adaptive_rho was already set to false, so the OSQP block should already be deterministic, or at least more deterministic w.r.t. to the use of OSQP with adaptive_rho enable (see https://github.com/google/osqp-cpp#faq).

toolbox/library/src/OSQP.cpp
```diff
@@ -350,11 +366,21 @@ bool wbt::block::OSQP::initialize(BlockInformation* blockInfo)
...
     if (!ok) {
         bfError << "Failed to get parameters after their parsing.";
         return false;
     }
+    
+    // Check the maximum number of Iterations
+    if (maxIter > 0){
+       pImpl->maxIterations = maxIter;
+    } else {
+       bfError << "The maximum number of iterations is not valid.";
```
traversaro, line 382: Probably it is helpful to either `return false` or explain that we are going to use the maxIter value of 1 ?

matlab/library/WBToolboxLibrary_repository.mdl
```diff
@@ -1,14 +1,14 @@
 Library {
-  Name			  "WBToolboxLibrary_repository2"
+  Name			  "WBToolboxLibrary_repository"
```
traversaro, line 2: @VenusPasandi did you renamed explicitly the library? Is that intentional?

traversaro:
I do know if https://it.mathworks.com/help/simulink/ug/make-backward-compatible-changes-to-libraries.html may be useful in this context.

toolbox/library/include/WBToolbox/Block/OSQP.h
```diff
@@ -41,6 +41,9 @@ namespace blockfactory {
  * | ParameterType::BOOL | 3 + Block::NumberOfParameters | 1 | 1 | "UseUb"         |
  * | ParameterType::BOOL | 4 + Block::NumberOfParameters | 1 | 1 | "ComputeObjVal" |
  * | ParameterType::BOOL | 5 + Block::NumberOfParameters | 1 | 1 | "StopWhenFails" |
+ * | ParameterType::BOOL | 6 + Block::NumberOfParameters | 1 | 1 | "AdaptiveRho"   |
+ * | ParameterType::BOOL | 7 + Block::NumberOfParameters | 1 | 1 | "Polish"   |
+ * | ParameterType::BOOL | 8 + Block::NumberOfParameters | 1 | 1 | "MaxIterations"   |
```
diegoferigo, line 46: super picky nit

```suggestion
 * | ParameterType::BOOL | 7 + Block::NumberOfParameters | 1 | 1 | "Polish"        |
 * | ParameterType::BOOL | 8 + Block::NumberOfParameters | 1 | 1 | "MaxIterations" |
```

toolbox/library/include/WBToolbox/Block/OSQP.h
```diff
@@ -58,8 +61,7 @@ class wbt::block::OSQP final : public blockfactory::core::Block
     bool parseParameters(blockfactory::core::BlockInformation* blockInfo) override;
     bool configureSizeAndPorts(blockfactory::core::BlockInformation* blockInfo) override;
     bool initialize(blockfactory::core::BlockInformation* blockInfo) override;
-    bool
-    initializeInitialConditions(const blockfactory::core::BlockInformation* blockInfo) override;
+    bool initializeInitialConditions(const blockfactory::core::BlockInformation* blockInfo) override;
```
diegoferigo, line 64: please run clang-format before merging (refer to https://github.com/robotology/wb-toolbox/blob/master/.github/CONTRIBUTING.md#code-style), should also solve an old issue https://github.com/robotology/wb-toolbox/issues/216

toolbox/library/src/OSQP.cpp
```diff
@@ -28,8 +28,6 @@
 
 using namespace blockfactory::core;
 
-const unsigned MaxIterations = 100;
-
```
diegoferigo, line 32: as a comment, as discussed in a meeting, this variable was never used in this block and it was here for copy-and-paste from the qpoases block

toolbox/library/src/OSQP.cpp
```diff
@@ -338,6 +352,8 @@ bool wbt::block::OSQP::initialize(BlockInformation* blockInfo)
     // PARAMETERS
     // ==========
 
+    int maxIter = 1;
```
diegoferigo, line 355: Maybe initializing to an invalid number would ensure that the parameter is correctly read from the mask

```suggestion
    int maxIter = -1;
```

diegoferigo (approved):
Added my review, nothing super important but nice to have. Approving in any case.

> The only doubt is how to handle the backward compatibility.
> [...]
> Now that we changed the mask of the block, what happens to models that were using the old OSQP block without this parameters? They will fail? They will use the default values of the new parameters? If they fail, what will be the error message?

It's long time I don't develop on Matlab/Simulink and I don't remember exactly, but I think that as long as the simulink mask initializes these variables so that their values match the previous hardcoded numbers, the behaviour of existing simulink models should not be affected.

matlab/library/WBToolboxLibrary_repository.mdl
diegoferigo: I have no memory of this but the 2 could have been there for showing simultaneously the mdl (developement version with newer matlab version) and slx libraries (deployed version, compatible with old matlab version).

In view of this, please double check that the committed simulink libraries are exported with the correct simulink versions (that I don't remember where we specified it).

toolbox/library/src/OSQP.cpp
VenusPasandi: I think using the default value of 1 is a nice idea. Please see a3fa2f82b01ac457184622ebbd5fb40fac3baa1a

diegoferigo: I agree with @traversaro that is better to make the block fail rather than applying something the user has not explicitly configured (example: I set -2 and magically the block sets 1. I'd prefer a failing simulation that tell me that I have to set an integeter > 0). The semantics of `bfError` is the scary red warning in the simulink GUI, and the model should fail in this case.

matlab/library/WBToolboxLibrary_repository.mdl
VenusPasandi: I have no idea about this extra "2". I didn't change the name of the library.

As I checked, previously it was "WBToolboxLibrary_repository", but then it changes to "WBToolboxLibrary_repository2" in the commit a1b81001950dd1768b54d8cc9b4b1659909aea25.

>> In view of this, please double check that the committed simulink libraries are exported with the correct simulink versions (that I don't remember where we specified it).

Yes, it is correct. The exported library is in R2016b.

toolbox/library/src/OSQP.cpp
diegoferigo: Furthemore, as replied in https://github.com/robotology/wb-toolbox/pull/234#pullrequestreview-1145947757, choosing a default value for the number of iterations that differs from the previous state (that was undefined, does OSQP has a documented number of iterations when not set?) would make this PR not backward compatible.

I suggest the following solutions:

- If OSQP has a documented number of iterations that is selected when the user does not specify it (-> what we were doing before this PR), use this number as default
- Othetwise, initialize in the mask an invalid number of iterations (e.g. -1) and do not call the `setMaxIteration` if it's less than 1. In this case, we can notify the user in the GUI with a message with low verbosity (lower than warning)

matlab/library/WBToolboxLibrary_repository.mdl
diegoferigo: Then ok, I guess that removing the 2 is ok, no idea why in that PR it was changed

toolbox/library/src/OSQP.cpp
VenusPasandi: Actually, I think we don't need to check the validity of the maxIterations in the C++ library, because in the Simulink GUI, 
- I defined an initial value (which is equal to 100 but I think we need to set it to 4000 for backward compatibility),
- I defined some constraints for the maximum iteration parameter, so in reality, the user can only choose a positive real number otherwise, the Simulink GUI generates an error.

So, I think we could remove this check from the C++ side.

VenusPasandi:
I cleaned the PR and merged all the commits in one clean commit.

toolbox/library/src/OSQP.cpp
```diff
@@ -350,12 +366,24 @@ bool wbt::block::OSQP::initialize(BlockInformation* blockInfo)
...
     if (!ok) {
         bfError << "Failed to get parameters after their parsing.";
         return false;
     }
 
+    // Check if the maximum number of Iterations read correctly from Simulink GUI
+    if (maxIter > 0){
+       pImpl->maxIterations = maxIter;
+    } else {
+       pImpl->maxIterations = 1;
```
diegoferigo, line 382: this line can now be removed, right?

VenusPasandi: Done

toolbox/library/include/WBToolbox/Block/OSQP.h
VenusPasandi: Done

diegoferigo:
Sorry @VenusPasandi, last thing. I was checking https://github.com/robotology/wb-toolbox/issues/231. Since we're updating this block and while waiting @traversaro's green flag, would you mind to replace the deprecated call with the new one?

gabrielenava:
>The only doubt is how to handle the backward compatibility. For this, I probably need the help of some Simulink expert (perhaps @gabrielenava @diegoferigo can be help me here). Now that we changed the mask of the block, what happens to models that were using the old OSQP block without this parameters? They will fail? They will use the default values of the new parameters? If they fail, what will be the error message?

>It's long time I don't develop on Matlab/Simulink and I don't remember exactly, but I think that as long as the simulink mask initializes these variables so that their values match the previous hardcoded numbers, the behaviour of existing simulink models should not be affected.

>@traversaro generally int he Simulink, there is no issue with backward compatibility unless,
>
> we change the number of inputs/outputs of a block,
> we add a parameter to the mask which is not initialized.
> In this PR, we don't change the input/output of the block and we initialized the three parameters that we added to the mask. So, I think this PR is backward compatible.

Agree with both  comments, I don't think there should be issues. In any case in IRonCub we use the OSQP block and `WBTollbox` in `master`, I can test it (even after merging the PR, so I will not block it) with the iRonCub controller and verify that our hypotheses are correct

traversaro:
Ok, you seem quite confindent on Simulink robustness to adding parameters. Then, for me we can merge once you add a description on this change in https://github.com/robotology/wb-toolbox/blob/master/CHANGELOG.md @VenusPasandi .

VenusPasandi:
> Sorry @VenusPasandi, last thing. I was checking https://github.com/robotology/wb-toolbox/issues/231. Since we're updating this block and while waiting @traversaro's green flag, would you mind to replace the deprecated call with the new one?

It is done in f7fbf5bfc78248a35c148d10fe40fc8c234e3faa

VenusPasandi:
> Ok, you seem quite confindent on Simulink robustness to adding parameters. Then, for me we can merge once you add a description on this change in https://github.com/robotology/wb-toolbox/blob/master/CHANGELOG.md @VenusPasandi .

The changelog is updated with the new changes.

traversaro:
Thanks!

PR #242, opened 2024-02-05
Debug RelativeTransform

traversaro:


LoreMoretti:
In https://github.com/ami-iit/element_qp-reactive-control/issues/75, we are using the `forward kinematics` block.

So maybe we should add some debug prints to https://github.com/robotology/wb-toolbox/blob/traversaro-patch-4/toolbox/library/src/ForwardKinematics.cpp.

traversaro:
> In [ami-iit/element_qp-reactive-control#75](https://github.com/ami-iit/element_qp-reactive-control/issues/75), we are using the `forward kinematics` block.
> 
> So maybe we should add some debug prints to https://github.com/robotology/wb-toolbox/blob/traversaro-patch-4/toolbox/library/src/ForwardKinematics.cpp.

I see, then we need to add some print around https://github.com/robotology/wb-toolbox/blob/traversaro-patch-4/toolbox/library/src/ForwardKinematics.cpp#L216 .

PR #244, opened 2024-09-24
Ensure that the OSQP report as status output not the return value of solveProblem(), but the output of solveProblem()

DavideBarbone:
As per title this PR contains the fix for the solver status of the Simulink OSQP block.

### DIFF
--- a/toolbox/library/src/OSQP.cpp
+++ b/toolbox/library/src/OSQP.cpp
@@ -599,7 +599,8 @@ bool wbt::block::OSQP::output(const BlockInformation* blockInfo)
     }
 
     // Set status
-    if (!statusSignal->set(0, double(solveReturnVal))) {
+    OsqpEigen::Status status = pImpl->sqSolver->getStatus();
+    if (!statusSignal->set(0, double(status))) {
         bfError << "Failed to set status signal.";
         return false;
     }


traversaro:
Technically speaking this is a change of behavior, but it actually permits to have a meaningful signal coming out of the status output, so I guess it is a fair change. I will nevertheless bump the minor version instead of the patch in the new wb-toolbox release to make that clear.

To provide more detail on the change: before this PR, the `status` output of the `OSQP` block returned the return value of `solveProblem`, see https://github.com/robotology/osqp-eigen/blob/v0.8.1/include/OsqpEigen/Solver.hpp#L123 . This method returned the `OsqpEigen::ErrorExitFlag`, that however returns OSPQ errors (see https://github.com/robotology/osqp-eigen/blob/b9ce7391350e0f1c71e49f7574a8785fb617b5e6/include/OsqpEigen/Constants.hpp#L43-L57), and not the solver exit status, that is instead contained in https://github.com/robotology/osqp-eigen/blob/b9ce7391350e0f1c71e49f7574a8785fb617b5e6/include/OsqpEigen/Constants.hpp#L23-L38 and returned by getStatus, see https://github.com/robotology/osqp-eigen/blob/b9ce7391350e0f1c71e49f7574a8785fb617b5e6/include/OsqpEigen/Solver.hpp#L173 . So this PR changes the logic to ensure that `status` return the output of `getStatus`. 

fyi @mfussi66 @Gio-DS

traversaro (approved):

traversaro:
Probably I merged this a bit in a rush, anyhow @gabrielenava you may want to have a look to this change.

PR #247, opened 2024-11-09
Do the changes to support BlockFactory 1.0.0 and bump version to 6.0.0

traversaro:
This PR does the changes required to support (and also require) BlockFactory 1.0.0 . See https://github.com/robotology/blockfactory/pull/87 for more information about BlockFactory 1.0.0 .

As this is effectively a breaking change, the version number is bumped to 6.0.0 .


traversaro:
As the CI installs `blockfactory` from conda, it fails as in conda we still do not have the latest blockfactory 1.0.0 . I think it is safe to merge as I compiled locally.

PR #253, opened 2025-10-07
Add Eigen3::Eigen link to WBToolboxBase target

traversaro:
I have no idea how this worked before, but indeed `Eigen/Core` header is used in WBToolboxBase, so we need to link the `Eigen3::Eigen` target. Somehow this used to work in some way, but it stopped working with Eigen3 5.0.0.

xref: https://github.com/robotology/robotology-superbuild/issues/1902
xref: https://github.com/robotology/robotology-superbuild/issues/1903
xref: https://github.com/conda-forge/eigen-feedstock/pull/47


traversaro:
I see, before this, the build system used `EIGEN3_INCLUDE_DIR` variable, that is not supported anymore.

